Issue #49, opened 2015-03-12
Run_time in SS_output


allanhicks:
The list from SS_output sets Run-time and repfiletime as the same, and they are both taken from the same line in the Report file.  It seems to me that Run_time should be the time it took to minimize (two lines below the Start Time in the header of the Report file).  I hesitate to change this in case this was intended. If so, it would be nice to have the time to minimize as output.  

A simple change to line 700 would do it:
stats$Run_time <- paste(as.character(matchfun2("StartTime",2,"StartTime",2,cols=4:9)),collapse=" ")

taylori:
Haven't tested change by @awhitten but change is straightforward and makes lots of sense.
Thanks.

taylori:
Thanks to @melmonk for noticing vestigial Run_time occurrence in SS_html. Reopening issue and will add fix and then close again.

taylori:
I said I would reopen, fix, and then close this issue. I seem to have made the fix but failed to close. Closing now.

PR #70, opened 2015-09-02
Update model 'StartTime' in SS_html

awhitten:
In relation to issue #49, @taylori made change to name of model `Run_time`, to `RunTime`, and replaced old `Run_time` with `StartTime`. `SS_html.R` still referred to `Run_time` so returned NULL and left the HTML 'Home' page empty. This simple fix updates the reference for 'Starting time of model:' to the new `replist$StartTime`.

Suggest also that issue #49 can now be closed. :)


awhitten:
I tested SS_html as part of SS_plots with png=TRUE (by default html=png), and all seems to be working fine. Cheers.

PR #78, opened 2015-11-06
Add a second plot of model data that shows the relative precision within a data type

colemonnahan:
I think this is ready to be implemented in master based on discussion in this commit: https://github.com/r4ss/r4ss/commit/2de27d95f2809392fdbb31794454832482cfe6a7#commitcomment-14227098

I've updated the branch to have circles be proportional to area instead of diameter and included a more informative HTML caption. Attached is a screen shot of the new plot in a browser.

@taylori If there's a set of models you use to test new functionality I could run this code by those. I've only tested this on two very similar models. 

capture

### DIFF
--- a/R/SSplotData.R
+++ b/R/SSplotData.R
@@ -36,6 +36,8 @@
 #' @param verbose report progress to R GUI?
 #' @param datasize Add second data plot whose circles are proportional
 #' to either catch or relative uncertainty? Produced as data_plot2.png.
+#' @param maxsize The size of the largest bubble in the datasize
+#' plot. Default is 1/2.
 #' @author Ian Taylor, Chantel Wetzel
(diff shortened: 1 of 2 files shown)


colemonnahan:
@taylori I started a new branch to work on some wtatage plots and somehow pulled these changes into that one (https://github.com/r4ss/r4ss/commits/wtatage_2gender). I'm not sure what will happen if you pull this in and then pull that other branch in as well. It may be safest to ignore/delete this request and these changes will be implemented later from the other branch.

I mean this from a perspective of keeping the git workflow as simple as possible, because honestly I don't know how to fix complex branch conflicts.

taylori:
OK, will ignore current request for now to keep merging simple.

PR #80, opened 2015-11-19
Small fixes for CRAN compatibility

seananderson:
I made a number of edits to pass `R CMD check --as-cran` on the current stable version of R and the current version of R-devel.

There are a couple files where I think the Windows/Unix line ending characters made it look like I edited every line. I added a `.gitattributes` file, which I think should fix this in the future. At least it should for new files.

Note that in the development version of R you now have to explicitly import functions from packages that ship with R although you don't have to declare them in the DESCRIPTION file.


taylori:
Thank you @seananderson !
Are any further steps needed to allow us to submit this to CRAN to finally replace the long out-of-date version there? I would increment the version number to 1.24 and update the date, but otherwise does it look OK?

seananderson:
I don't remember there being any other issues, but before the package is submitted:
-  create a `cran-comments.md` as in https://github.com/ss3sim/ss3sim/blob/master/cran-comments.md and edit as needed; `devtools::release` will use this
-  create a NEWS.md file if you are going to to indicate any changes since the last time. I don't think this is required.
-  read through https://cran.r-project.org/web/packages/policies.html
-  send a copy to the  Windows build machine `devtools::build_win()`, which will also give you one last check on the development version of R
-  we need to coordinate the submission of ss3sim and r4ss so that they go in at the same time. I know that ss3sim requires the development version of r4ss. I can't remember if there is a problem the other direction as well. If there is, and it shows up as a NOTE, then it should be mentioned in the `cran-comments.md`

PR #83, opened 2016-01-08
Halibut

colemonnahan:
@taylori 

As you know I've been helping Ian S. with the halibut assessment. Some of his tasks for me were to update/modify r4ss in a way that works for him but should also work for others. I've made all of these changes on the halibut branch:

https://github.com/r4ss/r4ss/commits/halibut

He thinks that a few of these in particular will be of interest to the SS community and that it might be worth trying to get them into the master branch.

Big changes are:

1. Added a new data plot whose circles are proportional (data_plot2.png). Turned on by default, circle sizes and transparencies controlled through top level argument in SS_plots.
2. Redid the empirical weight-at-age plots so they have a legend instead of text (although you can still turn that on; bio4_weightatsize.png)
3. Sex ratio plots for age composition data. Not working for any other data type b/c the halibut assessment doesn't have it, but should be extendable to length easily. See SSPlotSexRatio.R and make_multifig_sexratio.R which were borrowed heavily from SSPlotComps and make_multifig. (sexratio_age-flt10mkt0.png). Admittedly this one is the most "rough" in that it would need further testing.

These work with two halibut models, but these only really differ in the length of the time series, not in any big structural way. I don't have a lot of other models to test them, but I think the new functions should be close.

Up to you which, if any, of these updates to include Ian thinks the sex ratio ones in particular would be useful for @allanhicks.

bio4_weightatsize
data_plot2
sexratio_age_flt10mkt0


allanhicks:
Thanks Cole,
I look forward to trying some of these out with hake. I'll have to wait on
investigating the sex ratios until the hake assessment is complete, though.
Allan

Issue #48, opened 2015-02-21
convert tcltk functions to shiny?


iantaylor-NOAA:
Functions `movepars`, `selfit`, and `selfit_spline` all use tcltk package. The syntax is arcane and the package has some installation issues on some computers. Shiny may be an easier way to go for this type of function.

allanhicks:
I started working on a shiny implementation for selectivity. How would you like me to upload this example? Shiny requires a folder containing server.R and ui.R, plus the call to invoke these. I'm not sure if we can have a single ui.R and server.R to invoke all the things we want, but that may be difficult to design the ui.

One thing I was thinking is to create a branch, and then a folder called Shiny in the top level (alongside the R and man directories). This would allow us to test Shiny for now and introduce it to the R library when ready.  Let me know if you have other ideas.

taylori:
Allan,
Thanks for working on this.

It might be worth consulting with @seananderson on this as he understands better the folder structure required for R packages. It might be better to put the shiny-related stuff into an "extra" or "inst" folder as I believe that these are standard R package folders (at least they are present in https://github.com/ss3sim/ss3models which Sean just reorganized). In a branch it won't matter, but if it's easy to start out on the right path, that could save time later on.
-Ian

seananderson:
`inst` is a standard package folder that is installed with the package and is accessible to the user through, say, `system.file()`. You can put any folder you want inside there as long as it doesn't match a folder name that already has special properties (`doc` is an example of that, since it houses the rendered vignette files).

`extra` isn't a standard folder, but it's an example of how you can add any folder you want at the root level as long as you add it to the `.Rbuildignore` file. Otherwise, `R CMD build` will issue a warning about a nonstandard folder. This, of course, means that it's not accessible in the loaded package.

I don't have much experience with Shiny yet, but one option that seems to work well for cases where the level of website customization is minimal and the output is more report-like is to [embed the Shiny code into an R Markdown document](http://rmarkdown.rstudio.com/authoring_embedded_shiny.html). This saves you from having to deal with separate `.R` files and takes care of the web layout side of things. You'd render that into an HTML file with `rmarkdown::render()`.

I imagine that eventually you'd have code in functions within the `R` folder that when called renders an HTML page with Shiny embedded. Not sure where it would be best to house any template files you need, but somewhere in the `inst` folder might make sense.

allanhicks:
[...]

PR #109, opened 2016-06-30
Require sudo for Travis-CI

seananderson:
Try this? See https://docs.travis-ci.com/user/ci-environment/ 
issue #48

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,8 @@
 
 language: c
 
+sudo: required
+
 before_install:
   - curl -OL http://raw.github.com/craigcitro/r-travis/master/scripts/travis-tool.sh
   - chmod 755 ./travis-tool.sh


taylori:
Thanks @seananderson 
Let's see what happens.

taylori:
That seemed to work. There remains a warning:
```
* checking whether package ‘r4ss’ can be installed ... WARNING
Found the following significant warnings:
  Warning: no DISPLAY variable so Tk is not available
See ‘/home/travis/build/r4ss/r4ss/r4ss.Rcheck/00install.out’ for details.
```
But it says "passed" and gives the green light, which is what was most important to me so that when something else breaks we can see it.

Thanks for your help @seananderson.

seananderson:
Great. I think the warning is just noting that there is no screen (and hence display variable) attached to the headless server and so the Tk GUI won't work. That shouldn't matter here. Now you can add the badge to the readme.md file to display your pride (or shame) in the build status if you want. 

```
[![Build Status](https://travis-ci.org/r4ss/r4ss.png?branch=master)](https://travis-ci.org/r4ss/r4ss)
```

PR #121, opened 2016-10-17
Use hidewarn for 3.24f tag output warning

iagomosqueira:
This particular message could be hidden if hidewarn=TRUE. Probably some others too, but this one is hitting me right now.


taylori:
Thanks @iagomosqueira. Feel free to add this option to turn off any other warnings that you wish.

PR #142, opened 2017-04-06
Minor changes

kellijohnson:
@taylori I made a few changes based on `devtools::check()` giving errors when I was building the package. Before you merge can you double check e3fea00 and 2920e91 to make sure that I did not do something stupid and c5ca64d  to make sure that I did not do something that you don't agree with.


taylori:
The bug fixes are great. Thanks for doing this.
However, I want to leave tcltk out of the imports, so I think it would be better to revert c5ca64d and live with the error in the package check until the functions that depend on tcltk can be replaced (as noted in issue #48). The problem, pointed out by @yukio-takeuchi is that it can be problematic to install tcltk on Macs as well as machines like linux servers and that don't have a GUI.

kellijohnson:
Sounds good, and which is exactly why I did not push everything to master. Still trying to get up to speed.

taylori:
No problem. Some of the discussions that led to the changes were offline anyway.

However, I don't know how to merge a subset of a pull request. Is it necessary to change the pull request or can I just pick the commits that I want to merge?

kellijohnson:
[Cherry-pick](https://mattstauffer.co/blog/how-to-merge-only-specific-commits-from-a-pull-request)

taylori:
cherry-pick worked although in this case it would probably have been easier to merge the full pull request and then undo the tcltk thing manually.

PR #262, opened 2019-01-10
Respect age_selex_types[,4] on pattern 17

heavywatal:
The number of parameters for age selectivity pattern 17 depends on the 4th column ("Special") of `age_selex_types` matrix. See page 68–72 in "User Manual" version 3.24f.


taylori:
Looks good. Thanks for the fix. A similar change should be made to SS_readctl_3.30 at some point in the future.

heavywatal:
Thank you for merging this, @taylori. I have made another PR for 3.30. #263

PR #263, opened 2019-01-31
Apply #262 to SS_readctl_3.30 as well

heavywatal:
All the variables are assumed to be preserved between two versions.


taylori:
Looks good. Thanks @heavywatal.

Issue #138, opened 2017-03-31
develop SS_readctl_3.30 and SS_writectl_3.30


iantaylor-NOAA:
Developing read/write functions for control files seemed to be a very difficult task but @yukio-takeuchi managed to get it done. At this point, those functions only work for SS version 3.24. Developing functions for 3.30 isn't a top priority until users of 3.24 versions want to upgrade. This issue is to note that the 3.30 versions of these functions haven't yet been developed and to track progress if/when we choose to do so.

kellijohnson:
As written `SS_readctl_3.24` and `SS_readctl` do not match in their argument names. I am uncertain which one to change, but `Nfleet` is wrongly mapped to `Nfish` and `Nsurveys` is being pseudo mapped to `Nsurv` through partial matching. I suggest that the discrepancies be cleared up such that both functions use names that are used throughout `r4ss` (i.e., `Nfleet` and `Nsurveys`). I am happy to do this, I just wanted to make sure that it makes sense to @taylori and @yukio-takeuchi

taylori:
Thanks for catching this @kellijohnson.

Since SS_readctl_3.30 hasn't yet been written (@yukio-takeuchi has volunteered to work on it starting in September), I think the few users of these files (including @iagomosqueira) have probably just called SS_readctl_3.24 directly rather than the wrapper function SS_readctl which is out of date so the change should take place there,

The consistency question is a big enough problem that I will post a separate issue (probably will be #140) about it.

yukio-takeuchi:
I agree that there is some inconsistency	 of arguments names in between SS_readctl_3.24 and SS_readctl.
I remember that those inconsistencies happened when I enhanced SS_readctl_3.24 to cover more options.
Until when SS_readctl_3.30 is available ( I hope, I can work on it late this year), I also suggest to call SS_readctl_3.24 directly.

nklaer:
I have SS_readctl_3.30 working for the few example files that I have available, based on Yukio's 3.24 code and called via SS_readctl. I'll talk to Ian Taylor about steps for getting it into R4SS.

taylori:
@nklaer, thanks for the good news.
The formal git approach would be to commit the changes in a fork of the repository and then issue a pull request but I'm also perfectly happy to integrate them into the package myself if you just email me the new/modified files.

taylori:
Functions from @nklaer have been added in commit adeba65 to the `development` branch, available via 
```
[...]

PR #266, opened 2019-02-18
Fix two bugs in SS_readctl(use_datlist = TRUE)

heavywatal:
1. `N_CPUE_obs` length should be `Nfleet + Nsurveys`.
2. `fleetnames` should be stored in `ctllist`.


taylori:
Thanks for these changes.

I note that the read of the Q_setup section in SS_readctl_3.30 puts things in the format used by SS version 3.24. If/when an SS_writectl_3.30 function is written in the future (need noted in issue #138), this section should be revised to store the Q_setup in the format used by control files in SS 3.30.

Issue #253, opened 2018-11-05
PType column in output of SS_readctl_3.24


iagomosqueira:
SS_readctl_3.24 now adds a column named PType tp the SR, MGParms and initial_F data.frames. Is this correct for 3.24? A ctl file loaded and then saved fails to run in 3.24 unless I delete this column.

taylori:
Thanks for reporting this. It looks like the PType was added by Neil Klaer as part of all the improvements he made back in May (in commit adeba650f2a8ca7427e714b3d7bc1893a5c75100), but Neil was focused on reading files and clearly the link to SS_writectl_3.24 was not tested.

Without knowing how PType is being used, I could see two easy options to fix this that hopefully wouldn't break anything. We could add a switch to SS_readctl_3.24 that turns on or off the addition of this column.

Or even easier would be to add the following to the printdf function in SS_writectl_3.24 (which would go before line 118: https://github.com/r4ss/r4ss/blob/master/R/SS_writectl_3.24.R#L118:
```
        if("PType" %in% names(dataframe)){
          dataframe <- dataframe[ ,names(dataframe)!="PType"]
        }
```

iagomosqueira:
That should work, sure. Although code that manipulated the ctl list is also failing due to the extra column. My temporary fix was to drop PType from the various cbind calls, like line 513

```
ctllist$MG_parms_seas<-cbind(ctllist$MG_parms_seas,PType)
```

This could be controlled by an extra argument. ptype=TRUE. Will try to create a pull request.

taylori:
Having the option to exclude the column seems to be the cleaner solution anyway, so pull request for that approach would be great. Thank you.

PR #267, opened 2019-02-19
Exclude PType column from MG_parms in SS_writectl output

heavywatal:
The column is added to `MG_parms` during `SS_readctl` process, and should not
be written to control files. Otherwise the roundtrip of read-write
is impossible.

https://github.com/r4ss/r4ss/blob/871255ec83022df6594225a308c37c15d6da0c6d/R/SS_readctl_3.24.R#L481-L482

### DIFF
--- a/R/SS_writectl_3.24.R
+++ b/R/SS_writectl_3.24.R
@@ -115,6 +115,7 @@ SS_writectl_3.24 <- function(ctllist,outfile,overwrite=FALSE,verbose=TRUE,
     }
     if(!is.null(dataframe)){
       if(header){
+        dataframe$PType <- NULL
         names(dataframe)[1] <- paste("#_",names(dataframe)[1],sep="")
         writeLines(paste(names(dataframe),collapse="\t"),con=zz)
       }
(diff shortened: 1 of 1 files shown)


heavywatal:
My second commit fixes `printdf` in `SS_writectl`, and this change is applied to many other data.frames with `PType`:

https://github.com/r4ss/r4ss/blob/871255ec83022df6594225a308c37c15d6da0c6d/R/SS_readctl_3.24.R#L507
https://github.com/r4ss/r4ss/blob/871255ec83022df6594225a308c37c15d6da0c6d/R/SS_readctl_3.24.R#L519
https://github.com/r4ss/r4ss/blob/871255ec83022df6594225a308c37c15d6da0c6d/R/SS_readctl_3.24.R#L571
https://github.com/r4ss/r4ss/blob/871255ec83022df6594225a308c37c15d6da0c6d/R/SS_readctl_3.24.R#L632

taylori:
Note that a fix to this issue already in place (discussed in issue #253) is to exclude the "PType" column at the SS_readctl step via the `ptype=FALSE` input added by @iagomosqueira.

However, even for a user who wants access to the PType column within R or forgets to remove it using `ptype=FALSE`, I can't imagine a case where PType would be wanted in the control file written at the SS_writectl step. So this change makes sense to me.

PR #268, opened 2019-02-21
Fix duplication check in SS_readctl

heavywatal:
`paste0()` with `sep=` argument works in an unexpected way. E.g., `paste0('a', 'b', 'c', sep=',')` ends in `'abc '`, not `'a b c'`. Duplication in lambda table has been checked with the strings concatenated by paste0, and it can result in false positive. E.g., rows (1,20,1,1,0) and (12,0,1,1,0) get marked duplicated.


taylori:
It's not clear to my why there would be duplicate rows in the lambda inputs in the first place, but the change in how they are filtered is definitely an improvement.

PR #275, opened 2019-03-20
Fix so a relative path for function input dir can be used with function SS_profile().

doering-kat:
Tested and SS models ran for both relative and absolute specifications for dir with this change. Before, models would not run if a relative dir was used.


taylori:
Looks good. Thanks @doering-kat !

PR #277, opened 2019-03-29
Development: fix working directory change in SS_profile

k-doering:
Small changes to SS_profile function:
1. Change a default value for model to match default value in function.
2, Also change so working directory always remains the same for the user regardless of when the function exits (on error, or completes).  Change inspired by code in ss3sim::ss3sim_base() (not sure how to give credit, sorry!)


k-doering:
Also, should note that I tested the changes, and seems to work as expected with both relative and absolute paths for `dir`

taylori:
Change looks great. Thank you @k-doering !

PR #290, opened 2019-05-02
Writedat3.3

kellijohnson-NOAA:
Changed code to read and write 3.30 dat files that work with stock synthesis.

The code was only checked on sablefish files. Nevertheless, they now at least work with one model where previously they didn't work with any model.

I also updated some documentation to make the package pass some checks. This is the last commit that I made. We can remove it from this pull request if you want.


iantaylor-NOAA:
@kellijohnson-NOAA feel free to delete branch if you aren't working on it further

PR #292, opened 2019-05-08
Biasramp3.30

kellijohnson-NOAA:
* Update r4ss::SS_fitbiasramp to work with 3.30 models
* Add some text-file output


iantaylor-NOAA:
SS_fitbiasramp wasn't working for control.ss_new comments associated with 3.30.13 vs 3.30.12. Thanks @kellijohnson-NOAA for catching that.

PR #295, opened 2019-05-15
fix: allow for matrix of Flimitfractions in forecast

kellijohnson-NOAA:
If mylist$Flimitfraction is negative, a matrix of years and limits
will follow Flimitfraction. I use -999 to find the end of the matrix.
Then, many i values are skipped over so that the matrix values are not
wrongly put into other named values.

The write forecast function then puts writes these numbers back to the
file if Flimitfraction is negative.

I don't predict that the changes will have any conflicts with other
versions of SS.

### DIFF
--- a/R/SS_readforecast.R
+++ b/R/SS_readforecast.R
@@ -96,6 +96,15 @@ SS_readforecast <-  function(file='forecast.ss', Nfleets, Nareas, nseas,
     mylist$BforconstantF <- allnums[i]; i <- i+1
     mylist$BfornoF <- allnums[i]; i <- i+1
     mylist$Flimitfraction <- allnums[i]; i <- i+1
+    if (mylist$Flimitfraction < 0) {
+      ii <- i
+      while (allnums[ii] != -999) ii <- ii + 1
+      mylist$Flimitfraction_m <- data.frame(matrix(allnums[i:(ii + 1)], 
+        ncol = 2, byrow = TRUE))
+      colnames(mylist$Flimitfraction_m) <- c("Year", "Fraction")
+      i <- ii + 2
+      remove(ii)
+    }
     mylist$N_forecast_loops <- allnums[i]; i <- i+1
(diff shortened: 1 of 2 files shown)


kellijohnson-NOAA:
Allow Flimitfraction == -1
read and write forecast doesn't work if Flimitfraction is negative. Fix allows for reading of the two-column data frame that follows Flimitfraction in the forecast file if Flimitfraction is negative. Only tested with 3.30.

iantaylor-NOAA:
Good catch, thank you @kellijohnson-NOAA !

iantaylor-NOAA:
Recent change to `SS_readforecast` didn't quite work for a file I just tested the following buffer settings due to terminator code of -9999 vs -999 expected in `SS_readforecast`. It turns out that SS will take any negative year as the terminator (https://vlab.ncep.noaa.gov/redmine/projects/stock-synthesis/repository/revisions/master/entry/SS_readdata_330.tpl#L2904), so there's no correct terminator to use (although -9999 is more standard, I think). 

I generalized the r4ss function to use any negative year as well in commit f930c54. I also added a testthat test for SS_readforecast and SS_writeforecast, although the simple model in the test doesn't include time-varying buffer on the control rule so this aspect won't be tested.

@kellijohnson-NOAA let me know if the revised function causes any issue for models you're working with. Also feel free to delete this branch if you wish.

```
-1 # Control rule target as fraction of Flimit (e.g. 0.75), negative value invokes list of [year, scalar] with filling from year to YrMax
# buffer values below based on Category 2 Sigma with P* = 0.45
#Yr    buffer
2019   1.0
2020   1.0
2021   0.874  
2022   0.865  
2023   0.857  
2024   0.849  
2025   0.841  
2026   0.833  
2027   0.826  
2028   0.818  
2029   0.810  
2030   0.803
-9999  0
```

kellijohnson-NOAA:
Sorry @iantaylor-NOAA . Good to know that any negative number will work. Your fix works with sablefish.

PR #302, opened 2019-05-23
Change readctl function to work when there are no variance adjustment parameters

lee-qi:


iantaylor-NOAA:
The readctl function didn't work for the 3.30 model I just tried (totally unrelated to this change), so I haven't tested the change, but it looks like it should work fine.

k-doering-NOAA:
@iantaylor-NOAA, I also had some issues - for me the root of the problem seemed to be the error @lee-qi just fixed, as well as the number of parameters for rec_dist depends on the rec_dist_method used in 3.30. The code currently does not reflet this. I fixed it for 1 method and could read a control file, but have not fixed it for all methods yet - hopefully will finish this tomorrow and send a pull request.

PR #305, opened 2019-05-24
Edit readctl

k-doering-NOAA:
This change fixes issues with reading the number of recruitment distribution params read in  MGparms, which varies by method. Also adds warnings when using recruitment distribution method 1 or recruitment method interactions, which should not be used in 3.30 but were possible to use  in early versions of SS 3.30. Finally, it cleans up some inconsistencies among the functions in terms of how variables are named.

It includes a fix to allow the control file to be read when no variance adjustment was used, but this was taken care of by @lee-qi in commit c9feed80892a7d94cab2e8ca0489c38ad69d7049 that is now incorporated in master and development branches.

Note that I created this branch off an older version of development (last commit fae607a1c3c63a78b20714dce9e492b5b60e7f5f), which is why it can't be automatically merged. If it makes more sense to, I can cancel this pull request, rebase the latest r4ss development branch onto edit_readctl, and then extend a new request.

### DIFF
--- a/R/SS_readctl.R
+++ b/R/SS_readctl.R
@@ -29,11 +29,16 @@
 #'  explicitly available in control file
 #' @param Nsurveys number of survey fleets in the model. This information is also not
 #'  explicitly available in control file
-#' @param DatFile read datfile list for additional information for version 3.30
-#' and above
 #' @param N_tag_groups number of tag release groups in the model.
 #' This information is also not explicitly available in control file.
 #' @param N_CPUE_obs numbere of CPUE observations.
+#' @param use_datlist LOGICAL if TRUE, use datlist to derive parameters which can not be
+#'  determined from control file
+#' @param datlist list or character. if list : produced from SS_writedat
+#'  or character : file name of dat file.
+#' @param ptype include a column in the output indicating parameter type?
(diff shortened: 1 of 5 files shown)


iantaylor-NOAA:
Thanks @k-doering-NOAA! All these changes appear to expand the range of models for which the readctl functions work with while also making some of the inputs more consistent with other functions in r4ss. 

As discussed in issue #138, I don't often use the read/write ctl functions and didn't think it would be worth the trouble to develop them. Luckily @yukio-takeuchi, @iagomosqueira, @nklaer, @heavywatal, and perhaps others who I'm forgetting took the time to develop these functions and make further improvements. As the ones who are using them most, those folks might want to look over the changes in this pull request and make sure the modified functions continue to work for your needs.

In commit dc6d6c8b353d04303d323186feff04868a2e461a I've tried to make sure those contributors mentioned above (now including @k-doering-NOAA) are included in the list of authors of these read/write control functions, also including @kellijohnson-NOAA and @chantelwetzel-noaa on the read/write data functions. And @iagomosqueira was missing from the larger list of contributors to r4ss in DESCRIPTION, so that's been corrected as well. I also incremented the r4ss version number to 1.35.2 and ran `devtools::document()` on to automatically update all the .Rmd files.

PR #307, opened 2019-05-31
fix: use convert.type in get.df for 3.30 data to get numeric values

kellijohnson-NOAA:
get.df in SS_readdat_3.30 was returning character values
for numeric values. I first noticed this because I was trying to manipulate
some composition data. The CPUE data didn't display this issue because
SS_readdat converted some of those characters to numeric values.
I tried for a more general fix that would simplify the code, rather than changing all values using as.numeric(as.character)) because as.data.frame() was not working.

convert.type is used by read.table to convert to logical, integer,
numeric, or complex and only if those that fail will it generate character
values (or factors if as.is=TRUE)

I checked against two 3.30 data files that the results of the resulting
list were the same. I also went back in the history of the development
branch to ensure that the age and length comps had always been imported
as character values and that a recent change didn't create this problem.


iantaylor-NOAA:
Thanks for dealing with this. I had noticed the non-numeric values but hadn't followed recent changes to these functions well enough to know whether it was an issue or not.

I suspect that we should use `type.convert` (not `convert.type` as in the pull request title) would be useful throughout `SS_output` where there are numerous loops over columns to convert things to numeric.

kellijohnson-NOAA:
Ahh darn, sorry about the title. Yes, I just recently discovered ```type.convert```, but I like it better than ```as.numeric()```. 
A quick findstr for as.numeric revealed other functions, as well as other calls within ```SS_readdat``` that I missed, that could be changed to ```type.convert```. Perhaps, once assessments are over we can do this given that things aren't currently "broken".

k-doering-NOAA:
I noticed the component in the output from ```SS_readdat()``` called ```fleetnames``` now is a factor instead of a vector of characters, I think as a result of the change to using ```type.convert```. 

This is the default behavior to interpret character vectors as factors with ```type.convert``` so it is probably true if there are any other vectors of characters that are manipulated with ```type.convert```. Vectors of characters can instead be the default by using the ```as.is = TRUE``` argument. @kellijohnson-NOAA and @iantaylor-NOAA do you think this argument should be added? 

P.S. The use of ```type.convert``` is really smart - I never knew about this function!

kellijohnson-NOAA:
Again, darn. I read the help file wrong and thought that the default of ```as.is``` was what we wanted, but @k-doering-NOAA I think you are right in that ```as.is=TRUE``` should be used instead. Good catch. I will push this straight to development without bothering @iantaylor-NOAA.

PR #309, opened 2019-06-12
Fix: Add bycatch only fleet to SS_read/write_3.30 funs

k-doering-NOAA:
Bycatch only fleet info previously was being grouped in incorrectly as catch data in the SS_read/write_3.30 functions.This now allows it to be read and written in the correct format.

Testing done on a model with several bycatch only fleets as well as a model without.

Please feel free to change column or variable names if desired. I tried to use terms from ```data .ss_new``` 3.30.13 default comments when feasible.


iantaylor-NOAA:
Looks good to me. I see that bycatch fleets is on the list of features we need to test against in the "List of models for use in testing" Google Doc.
@kellijohnson-NOAA and @chantelwetzel-noaa, you know these functions better than me, so please feel free to comment or make changes if you see a need.

PR #314, opened 2019-06-24
Make initial version of SS_writectl_3.30

k-doering-NOAA:
This feature branch adds an initial version of ```SS_writectl_3.30```, and creates a wrapper function ```SS_writectl``` that should be used to call ```SS_writectl_3.30``` and ```SS_writectl_3.24``` (as done with the ```SS_readdat``` and ```SS_writedat``` functions). 

Please note that there are still limitations for what ```SS_writectl_3.30``` can write: the biggest limitations are 1) ctl files with multiple platoons cannot be written, 2) growth options 4,5, and 8 cannot be written, and 3) time varying parameters (including devs, env links, and blocks) cannot yet be written. This will need to be implemented in the future, but for now, ```stop()``` with an informative error message is used when one of these situations is encountered. In some cases, reading control files using these options these will require changes to both ```SS_readctl_3.30``` and ```SS_writectl_3.30```.

Also to note:
- A new testthat script called test-control.R was created that adds new, simple automated tests for the read and write control functions. These tests were passing after all code changes were complete.
-```devtools::document()``` was run after code changes complete.
-```devtools::check()``` also ran, and some warnings were triggered, although it did not fail and the warnings were not in reference to any new parts in the code. For now, I did not make changes based on these warnings, but it would probably be a good idea to address them relatively soon.

For now, SS_writectl_3.30 and SS_writectl_3.24 are exported functions, but perhaps they should not be...feel free toedit.


iantaylor-NOAA:
Thanks for doing all of this @k-doering-NOAA!
The addition of the testthat test is really valuable and serves as a model for the steps we should all take more often.

k-doering-NOAA:
I also failed to mention that most of the code was really already written, thanks to the authors (@yukio-takeuchi , @nklaer , and  @heavywatal) of the SS_readctl and SS_writectl functions! I included their names as authors on the ```SS_readctl_3.3``` function, but I am not sure if this is appropriate or not. Authors and Ian, feel free to change if what I did was incorrect.

PR #315, opened 2019-06-29
Fix changepars

kellijohnson-NOAA:
Development in SS_parlines and SS_changepars. I think they are all okay, given they pass the new tests that I made, but (3) could be thought about more and better developed to work with (4).
(1) SS_parlines object ncols was not large enough for sablefish model, now it dynamically creates a large enough matrix.
(2) SS_parlines interpreted blocks_per_pattern as a parameter
(3) SS_changepars would reorder parameter names if unique() removed duplicated names causing the newvals to be out of order with the names. My changes here could potentially be further developed, but at least they remove the error that I found. Unfortunately, they still don't work if you want to use repeat.vals = TRUE and give it a vector of the same length as strings but have each individual string match multiple parameter names. Something to think about, but right now a stop message stops this.
(4) SS_changepars repeats all input vectors if length of the vector is one and repeat.vals = TRUE

### DIFF
--- a/R/SS_changepars.R
+++ b/R/SS_changepars.R
@@ -112,7 +112,10 @@ function(
   fullctlfile <- file.path(dir, ctlfile)
   ctl <- readLines(fullctlfile)
 
-  # check for valid input
+# check for valid input
+  inargs <- list("newvals" = newvals, "newlos" = newlos, "newhis" = newhis, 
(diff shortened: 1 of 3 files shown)


iantaylor-NOAA:
Looks good, thanks @kellijohnson-NOAA.

PR #318, opened 2019-07-05
enable read/write_forecast with fleet_relative_F == 2

mkapur:
I wrote a line into SS_readforecast which can deal with fleet_relative_F == 2 (retaining the correct order of mylist) and disabled the attendant stop() in SS_writeforecast . Worked reliably on my china rockfish iterations.

### DIFF
--- a/R/SS_readforecast.R
+++ b/R/SS_readforecast.R
@@ -100,8 +100,9 @@ SS_readforecast <-  function(file='forecast.ss', Nfleets, Nareas, nseas,
     mylist$Ydecl <- allnums[i]; i <- i+1
     mylist$Yinit <- allnums[i]; i <- i+1
     mylist$fleet_relative_F <- allnums[i]; i <- i+1
-    if(mylist$fleet_relative_F==2){
-      stop("SS_readforecast doesn't yet support option 2 for 'fleet relative F'")
+    if(mylist$fleet_relative_F==2){ ## START 35
+      mylist$vals_fleet_relative_f <- allnums[i:(i+Nfleets-1)]; i <- i+Nfleets
+      # stop("SS_readforecast doesn't yet support option 2 for 'fleet relative F'")
     }
     mylist$basis_for_fcast_catch_tuning <- allnums[i]; i <- i+1
     if(version==3.24){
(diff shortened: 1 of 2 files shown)


iantaylor-NOAA:
@mkapur Thanks for contributing this change. I was probably just too lazy to implement this when the function was first written many years ago, so it's great to have you finally fix this.
Can I request two minor adjustments to the pull request. First, the change to `SS_writeforecast` seems to need some additional code to write the new `$vals_fleet_relative_f` vector. Second, it's not a big deal, but from a workflow perspective, it's probably better to issue the pull request to the `development` branch rather than `master`.
Also, feel free to just delete the `stop()` lines rather than commenting them out as git keeps track of what was there before.

mkapur:
Thanks Ian -- and sorry about the wrong branch. I'm at the sablefish review
this week but will add this to my to-do list for next week.
It'll be good if we can get this merged at some point since it's a key
feature of the catch-only iteration function I'm developing.
MK

iantaylor-NOAA:
I just merged Development into Master (pull request #317) so merging this change to Master as well is fine. I saw another thing I'd like to change in the read/write forecast files related to differences between 3.24 and 3.30, so I'm happy to add the additional code to write the new $vals_fleet_relative_f.
Thanks @mkapur for your help with this.

iantaylor-NOAA:
Committed 53435f3 to `development` branch has some further changes to these files to try to get them to work with fleet relative F in version 3.30 as well as 3.24. I tested on one model of each type and they seem to now work.
@mkapur, when you have more time next week or in the future, please let me know if the modified files cause any problems with your catch-only iteration function.

mkapur:
Thanks @iantaylor-NOAA -- sorry, hope you didn't take my email to mean we needed to get this merged in a hurry, but thanks anyway for looking at it. I'll revisit next week to see if they are functioning as expected.
MK

iantaylor-NOAA:
One more complexity I just discovered thanks to @John-R-Wallace-NOAA is that the fleet_relative_F == 2 option was broken in SS version 3.30.07.01 used for some of the 2017 assessments (and perhaps all earlier 3.30 versions). So the read/write forecast functions as modified in the `development` branch of r4ss yesterday should work for SS version 3.24 and versions 3.30.08.04 and beyond that contain fleet_relative_F == 2. Earlier 3.30 versions would work with the r4ss functions only have they have If they have fleet_relative_F == 1, which they should since option 2 isn't working for those versions.

PR #328, opened 2019-07-25
Fixes to SS_readctl_3.30 and SS_writectl_3.30

k-doering-NOAA:
Small fixes, as noted here: https://github.com/r4ss/r4ss/issues/138#issuecomment-512586458 and https://github.com/r4ss/r4ss/issues/138#issuecomment-515111770. Also small changes in style and made verbose = FALSE actually not verbose. for SS_readctl_3.30.

### DIFF
--- a/R/SS_readctl_3.30.R
+++ b/R/SS_readctl_3.30.R
@@ -629,9 +629,9 @@ SS_readctl_3.30 <- function(file,verbose=TRUE,echoall=FALSE,version="3.30",
   
   ctllist<-add_df(ctllist,name="MG_parms",nrow=N_MGparm,ncol=14,
                     col.names=c("LO", "HI", "INIT", "PRIOR",
-                                "PR_type", "SD", "PHASE",
+                                "SD", "PR_type", "PHASE",
                                 "env_var","use_dev", "dev_minyr",
-                                "dev_maxyr", "dev_stddev", "Block", "Block_Fxn"),
+                                "dev_maxyr", "dev_PH", "Block", "Block_Fxn"),
                                 comments=MGparmLabel)
   
   ctllist$MG_parms<-cbind(ctllist$MG_parms,PType)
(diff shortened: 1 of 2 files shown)


k-doering-NOAA:
Also, I ran `devtools::check()` locally and got no errors, but the following warning and notes:
```
-- R CMD check results ---------------------------------------------------------------------------------- r4ss 1.35.3 ----
Duration: 6m 23.9s

> checking for unstated dependencies in examples ... OK
   WARNING
  'qpdf' is needed for checks on size reduction of PDFs

> checking dependencies in R code ... NOTE
  'library' or 'require' call to 'shiny' in package code.
    Please use :: or requireNamespace() instead.
    See section 'Suggested packages' in the 'Writing R Extensions' manual.

> checking R code for possible problems ... NOTE
  SS_writedat_3.30: no visible binding for global variable 'fleetname'
  make_multifig_sexratio: no visible global function definition for
    'qbeta'
  Undefined global functions or variables:
    fleetname qbeta
  Consider adding
    importFrom("stats", "qbeta")
  to your NAMESPACE file.

0 errors √ | 1 warning x | 2 notes x
```

I don't think these are from changes I made, but happy to fix if I am mistaken.

kellijohnson-NOAA:
I don't think any of those are from your changes. The shiny error and qpdf error have been there for a while.

iantaylor-NOAA:
@k-doering-NOAA, thanks for running `devtools::check()` which I don't do often enough, and bringing up these warnings, which I don't pay enough attention to sometimes.

I just made a few changes in commit 61460d8 to reduce the warnings: adding qbeta to the import list in r4ss-package.R (which causes NAMESPACE to get automatically updated via `devtools::document()` and replacing 
```
subset(d$bycatch_fleet_info, select = -fleetname)
```
in `SS_writedat_3.30` with
```
d$bycatch_fleet_info[,names(d$bycatch_fleet_info) != "fleetname"]
```
which is less compact, but is consistent with the rest of the code (since I never knew about `subset()` until very recently) and makes the warning go away. According to Hadley, it seems that you're stuck with the NOTE about that type of use of subset: https://stackoverflow.com/questions/4184117/r-cmd-check-notes-on-ggplot-subset-etc

PR #329, opened 2019-08-01
Fix for SS_writectl_3.30()  when N_lambdas = 0

k-doering-NOAA:
Small fix. Adds terminator line to output SS file when N_lambdas = 0 and also checks for input consistency. ran test-control testthat() automated tests, and read in/out ctl files with and w/o lambdas in the file.

Note I did not run devtools::check() locally b/c the changes were so minimal.


iantaylor-NOAA:
I haven't done further testing, but it looks good. Thanks for making this improvement.

Issue #251, opened 2018-10-17
"maxyr" input not fully implemented


taylori:
Thanks to Max Cardinale for flagging this. The `maxyr` input to `SS_plots` (and some associated functions it calls) changed the x-axis limits in some cases, but didn't remove the points beyond `maxyr`. In other cases, the input has no impact.

Commit 6b195e884c9bb306bb71938e9611770c53208c44 in the "development" branch makes `maxyr` work better for `SSplotTimeseries`.

Changes to other plots will take more time.
To get this change, Install the development branch using
```
devtools::install_github("r4ss/r4ss", ref="development")
```

akatan999:
Thanks Ian☺

Max

PR #332, opened 2019-09-09
Fix years plotted in rec deviations and ymax in recruitment time series

kellijohnson-NOAA:
2 observed problems
(1) using minyr in SSplotRecdevs was leading to the wrong years being plotted b/c it was subsetting the time series twice with goodyrs
(2) the ylim of SSplotTimeSeries(subplot = 11, minyr = x) was very large when I used the minyr argument, which was occurring because stdtable was not being truncated for the same years as the time series

I am pretty sure that I fixed (1), but the fix to (2) might not be the best. SSplotTimeSeries is more complicated and so I may have introduced problems for other things. I can easily revert this last commit and repush if you think that (2) should be fixed a different way and you just want to merge in the fixes for (1).

### DIFF
--- a/R/SSplotRecdevs.R
+++ b/R/SSplotRecdevs.R
@@ -103,7 +103,7 @@ SSplotRecdevs <-
 
       Yr <- c(recdevEarly$Yr,recdev$Yr,recdevFore$Yr)
       if(forecastplot){
-        goodyrs <- rep(TRUE,length(Yr))[Yr >= minyr & Yr <= maxyr]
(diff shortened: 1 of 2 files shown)


iantaylor-NOAA:
Thanks for making these fixes @kellijohnson-NOAA. As noted in issue #251 and elsewhere, the `minyr` and `maxyr` inputs have been slow to get fully implemented. In general `SSplotTimeseries` is a mess and your fixes look just fine as far as I can tell.

PR #333, opened 2019-09-13
changes to SSexecutivesummary

iantaylor-NOAA:
@chantelwetzel-noaa let me know if you'll have time to look through the changes which were associated with using this function for catch-only updates with an SS 3.24 model. Unfortunately I accidentally committed some of the reformatting changes (mostly indents and adding line breaks for extra-long lines) in the same commits as the substantive changes, so it's hard to look through the changes to see what the real differences are.

You can install the branch via 
`devtools::install_github('r4ss/r4ss', ref = 'SSexecutivesummary_changes')`
to confirm that the function still works for you.

### DIFF
--- a/R/SSexecutivesummary.r
+++ b/R/SSexecutivesummary.r
@@ -21,6 +21,8 @@
 #' based on the new sex
 #' specification (-1) in SS for single sex models. Default value is FALSE.
 #' TRUE will not divide by 2.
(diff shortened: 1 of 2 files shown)


iantaylor-NOAA:
Going ahead with merge. If problems with the change arise in the future we can open new issues.

PR #336, opened 2019-09-19
refactor: create SS_readwtatage

kellijohnson-NOAA:
Motivation ---
Needed to read in a weight-at-age file without having access to all the other
files associated with a stock synthesis run or take the time to read in the report
file or other data.

Changes ---
* Moved reading of weight-at-age data from SS_output to its own function
* Call SS_readwtatage from SS_output
* Use message rather than cat
* Returns a NULL if no file
* Fixed typos in verbose template

Potential problems ---
* Removed the terminator line of the weight-at-age data, i.e., Yr = -9999


iantaylor-NOAA:
Looks good. Thanks for doing this @kellijohnson-NOAA. There are surely additional improvements that could be made, such as figuring out why the original code replaced the header values with the vague comment "problems with header so simply manually replacing column names", but to achieve your immediate goal it's probably better to just do the refactoring without tinkering with the code.

kellijohnson-NOAA:
I only messed with stuff that I had to, such as `accuage`.

PR #346, opened 2019-11-07
feat(M_type): SS_ouput now reads M parameter and type

kellijohnson-NOAA:
"Natural_Morality" is a section of the report file where the
natural mortality value is specified per age. This output is
potentially different than the temporal Z at age matrix where
there is no value for the terminal age category and the next to
last age category is an approximation.
I also read in the M_type now as a list element.
This is needed for Claudio Castillo-Jordan's work.

I tried to match the code below for the growth parameters. 
Checked with devtools::check() with 0,0,0 notes, warnings, and errors.
Not entirely sure if this will break with older models because I only
checked it with 3.30.14.

### DIFF
--- a/R/SS_output.R
+++ b/R/SS_output.R
@@ -2022,6 +2022,14 @@ SS_output <-
                                        all(biology$Wt_len_F==biology$Fecundity),
                                        "biomass", "numbers")
 
+  M_type <- as.numeric(gsub(".*([0-9]+)", "\\1", 
+    rawrep[matchfun("Natural_Mortality"),2]))
+  M_Parameters <- matchfun2("Natural_Mortality",1,
+                            "Natural_Mortality",1+nrow(morph_indexing),
+                            header=TRUE)
+  returndat$M_type <- M_type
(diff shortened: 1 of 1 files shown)


kellijohnson-NOAA:
Oh geez, sorry you had to do so many modifications to get it to work. Glad I didn't push to development. One question I have is, should we change M_type to M_Type to match M_Parameters?

iantaylor-NOAA:
Don't worry @kellijohnson-NOAA it wouldn't have broken anything, just not achieved a usable table of M values for a rare subset of models.
Regarding capitalization, there is no consistent naming convention in r4ss.
There's also `fleet_type` and `discard_type` and `parameters` and `Growth_Parameters`.
The balance you've struck is just fine, so I went ahead with the merge.

While I have the floor, anyone reading this should consider tuning into my talk on r4ss at the CAPAM workshop on next generation stock assessment models at 9:30am New Zealand Daylight Time (2 hours from now). See http://capamresearch.org/Next-Gen-SAM-Remote for remote access info.

PR #347, opened 2019-12-03
Biasadjust

kellijohnson-NOAA:
SS_fitbiasramp was verbose even when verbose = FALSE and the output to the screen was clogging up ss3sim, which led to additional changes to the function to try and incrementally clean a few things up. Pull request includes
  1. fix up documentation from development to pass CRAN checks
  2. add verbose in front of cat()
  3. start using templates more in documentation, add some more info to SS_fitbiasramp
  4. more templates for arguments, these can be carried forward to other functions, but figured that would be better done directly in the development branch than here
  5. remove code that was commented out in SS_fitbiasramp

### DIFF
--- a/R/SS_fitbiasramp.R
+++ b/R/SS_fitbiasramp.R
@@ -50,16 +50,12 @@
 #' or "psoptim". If not equal to either of these, then optim is used.
 #' @param exclude_forecast Exclude forecast values in the estimation of
 #' alternative bias adjustment inputs?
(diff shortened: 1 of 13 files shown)


iantaylor-NOAA:
Thanks for these changes @kellijohnson-NOAA. In addition to making sure that `verbose` is working correctly, please feel free to change any use of `cat()` to `message()` as discussed at https://github.com/r4ss/r4ss/wiki/ideas-for-messages-and-warnings

PR #350, opened 2019-12-16
update SS-readdat_3.30.R to find 999 EOF marker

allanhicks:
It was looking for 999 without anything following it, but some users may have a comment after the 999. Using \\b finds the whole word. Tested with this code:

`tmp <- c("999 # End of data file","999","999#","xx","9999","99","999\t")`
`grep("^999$",tmp)`
`grep("\\b999\\b",tmp)`

### DIFF
--- a/R/SS_readdat_3.30.R
+++ b/R/SS_readdat_3.30.R
@@ -45,7 +45,7 @@ SS_readdat_3.30 <-
   }
 
   ###############################################################################
-  sec.end.inds <- grep("^999$", dat)
+  sec.end.inds <- grep("\\b999\\b", dat)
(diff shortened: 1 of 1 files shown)


iantaylor-NOAA:
Thanks @allanhicks for finding the problem and fixing it. I didn't know about the \b code for word boundary in regular expressions. I've merged this change to `master` into the `development` branch as well.

allanhicks:
No problem @iantaylor-NOAA and glad I could contribute 6 characters to the millions in r4ss. I just learned about \b because Ian Stewart decided to comment the line with 999 in the halibut assessment!

PR #355, opened 2020-01-06
Completed update to 3.30

nathanvaughan-NOAA:
Completed update to 3.30 for all scenarios including catch caps and allocations. Also converted to match read_ctl format with element, vector, and dataframe add functions.

### DIFF
--- a/R/SS_readforecast.R
+++ b/R/SS_readforecast.R
@@ -28,7 +28,7 @@ SS_readforecast <-  function(file='forecast.ss', Nfleets, Nareas, nseas,
   }
   
   if(version=="3.24"){
-    if(!exists(Nfleets) | !exists(Nareas) | !exists(nseas)){
+    if(!exists("Nfleets") | !exists("Nareas") | !exists("nseas")){
       stop('version 3.24 must include values for Nfleets, Nareas, and nseas. At least one of these is missing')
     }
   }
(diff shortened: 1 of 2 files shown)


iantaylor-NOAA:
Hi @nathanvaughan-NOAA,
Thanks for the extensive update to the read/write forecast functions.
Regarding the `argument 'nseas' is missing, with no default` error in Travis, I think that a good solution would be to provide default NULL inputs for all of `Nfleets`, `Nareas`, and `nseas`. I suppose it would also be good to add a warning for 3.24 models (which are the ones that require these inputs) along the lines of 
```
  if(forelist$SSversion==3.24 & is.null(nseas)){
    warning("Input 'nseas' can't be NULL for 3.24 models")
  }
```
Let me know if you want to make such changes or I can merge the pull request and do it myself.

Regardless, thanks again for making these functions better.
-Ian

nathanvaughan-NOAA:
Hey Ian,
  I think I have made the needed fixes now (I had a duplicate call to nseas
hiding further down the script). I also added a stop function for if those
values are missing in 3.24. Hopefully, it works this time.
Cheers,
Nathan

nathanvaughan-NOAA:
Ooops, my bad. I changed that to run a quick local test.

iantaylor-NOAA:
Looks good @nathanvaughan-NOAA. Sorry for the extra wrangling to get the test working. I think having NULL defaults also makes it more clear to SS version 3.30 users that an input isn't necessarily expected.

nathanvaughan-NOAA:
hahaha no worries, it was just me fixing errors in circles. I didn't
realize at first that it was the original function call making the error.
Got it working in the end though.

PR #364, opened 2020-01-21
feature: read in 2dAR selectivity parameters from control file

kellijohnson-NOAA:
Augment SS_readctl_3.30 to read the 2dAR selectivity parameters written
in the hake control file.

I did not create the dummy columns that were in the pseudo code previously
available.
Code still needs to be added to the write functions to handle these parameters.
Also, the internal function that I created might be better placed at the top of the read function. Not sure, but it was easier to read in the parameters using this internal function than the ones that were available.

I also have TMB code to estimate the parameters externally, but I am not sure where this code should be housed as it was provided in an email and requires TMB. Any ideas for how to store the code are welcome.

### DIFF
--- a/R/SS_readctl_3.30.R
+++ b/R/SS_readctl_3.30.R
@@ -1216,14 +1216,31 @@ SS_readctl_3.30 <- function(file,verbose=TRUE,echoall=FALSE,version="3.30",
                       comments = tmp_parlab)
   }
   #2DAR ----
-  ctllist<-add_elem(ctllist,name="Use_2D_AR1_selectivity") # Experimental facility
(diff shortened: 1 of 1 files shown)


iantaylor-NOAA:
Thanks @kellijohnson-NOAA. 
Let's talk about the TMB code after you get through the hake process and see how this all works for you.

kellijohnson-NOAA:
Thanks @iantaylor-NOAA. I emailed Haikun and he to share however we feel best facilitates its use in the broader community. @k-doering-NOAA do you think we can talk about it during an SS team meeting?

k-doering-NOAA:
@kellijohnson-NOAA that sounds like a great topic to discuss.

PR #372, opened 2020-02-07
Read Catch Bug Fix

nathanvaughan-NOAA:
Reading in Catch didn't have column labels which conflicted with the write file. Added names to match.


iantaylor-NOAA:
Hi @nathanvaughan-NOAA. It looks like the column names get added by the function `SS_readdat` (which calls `SS_readdat_3.30`). My understanding of the workflow that @k-doering-NOAA and @kellijohnson-NOAA have been working on is to have users only use `SS_readdat` and `SS_writedat` and remove the version-specific read/write files from NAMESPACE in the future. But I haven't followed the progress well enough to know if there would be problems with your proposed change or not, so will leave it to one of them to weight in on this. The conflict in the pull request appears to be only from you using a version of the file that had extra comments which have since been deleted.

nathanvaughan-NOAA:
awesome, thanks Ian. I didn't realize that. We will see what Kelli and Kathryn think. I was wondering how it made it through the checks. Have a great weekend.

k-doering-NOAA:
Thanks for the fix, @nathanvaughan-NOAA . While the approach @iantaylor-NOAA explained I think is consistent with the code I have contributed (I can't speak for all the other contributors who have put substantial more time than me into the read/write dat functions),  it is probably best that the `SS_readdat_3.30` and `SS_writedat_3.30` functions are directly usable until it is clear to all users that they should not be using them. If anyone has ideas on how to implement this, I think it is worth discussion.

I checked locally and the Travis CI passed, so this change seems compatible with the previous code. I will go ahead and merge into development.

iantaylor-NOAA:
Thanks @k-doering-NOAA. I agree that it makes sense to keep the individual version-specific functions able to work with each other until they get removed from NAMESPACE and we maybe better document the recommended workflow and reasons for not using those functions.

Issue #308, opened 2019-05-31
replace as.numeric with type.convert in SS_output and perhaps elsewhere


iantaylor-NOAA:
As noted in pull request #307, the `type.convert` function has some benefits over `as.numeric`, and could clean up many messy sections in `SS_output`.

In particular, blocks of code like
```
  # make values numeric
  for(icol in (1:ncol(parameters))[!(names(parameters) %in%
                                  c("Label","Pr_type","Status"))]){
    parameters[,icol] <- as.numeric(parameters[,icol])
  }
```
which occur in almost 40 places in `SS_output`, could be replaced by
```
  parameters <- type.convert(parameters, as.is = TRUE)
```
where the `as.is = TRUE` avoids conversion to factors.

iantaylor-NOAA:
Commit 75c6e03 in new branch `type.convert` replaces all the loops with `as.numeric()` in `SS_output()` to use `type.convert()` with no loop.
This reduces the function by 100 lines of code (out of about 3000) and makes it easier to read.

I used `dplyr::all_equal()` to compare the data frames created with the old approach to the new approach, which showed some differences in type for many of the tables. , where columns which were numeric but only contained integers are now stored as integers. NA values are now treated as logical rather than numeric. I don't believe any of these differences have any impact on the model results, but will keep these changes in the separate branch pending a bit more testing.

iantaylor-NOAA:
Running `SS_output()` for each of a set of 100 models took 2 minutes 23 seconds with the old approach and 2 minutes 20 seconds with the new approach. That tiny improvement might just be noise, but at least the `type.convert()` function noes not appear to be slower.

mkapur:
I updated my r4ss install and am now getting an error which may be related to this change.
I've confirmed that this error isn't showing up for the same model for a colleague who hasn't updated their r4ss in a while. 

I think the source is the call to `type.convert(df,...)` on in line 144 of [SS_readdat](https://github.com/r4ss/r4ss/blob/b90113ad2819bb5d183b0caf7b238f667863c87f/R/SS_readdat_3.30.R).

Let me know if I'm overlooking something obvious!
```
mod <- SS_output(path_to_model, 
                 printstats=FALSE, 
                 verbose = FALSE, 
                 hidewarn = TRUE, 
                 warn = FALSE,
                 forecast= FALSE
                 )

Error in utils::type.convert(df, as.is = TRUE) : the first argument must be of mode character
```

[...]

PR #374, opened 2020-02-07
Type.convert

iantaylor-NOAA:
Change from `as.numeric` within loops to `type.convert` as discussed in issue #308.


arni-magnusson:
Nice one. The `type.convert` is a great converter tool in R 3.5 onwards. The only drawback, which you have already identified, is the need to specify `as.is=TRUE` but that's a small price to pay :)

PR #375, opened 2020-02-10
fix indices plot in SSplotComparisons

kellijohnson-NOAA:
Problem = indices are not subset for endyr like other matrices. The hake assessment uses a group plot of five different comparisons (e.g., biomass, depletion, index) for bridging models and the different terminal year for the panels was driving Chris crazy. 
* truncate indices by endyrvec
* change the limit of the xaxis given the newly truncated
values or keep range if there is no endyrvec
* make the axes values pretty b/c lots overlap when you have
four digit year values with las = whatever makes the x axis parallel

One issue that I did not deal with because I am not sure how to, is the padding on the x axis. Other plots remove this padding automatically, but it makes it awkward with the index value because without the padding you cannot see the points at the min and max years very well. This pull request has the bare minimum of fixing of subsetting because I am not sure that fundamentally we want to remove the padding, i.e., `xaxs = "i"`.

### DIFF
--- a/R/SSplotComparisons.R
+++ b/R/SSplotComparisons.R
@@ -1283,7 +1283,7 @@ SSplotComparisons <-
     for(iline in 1:nlines){
       imodel <- models[iline]
       subset1 <- indices$imodel==imodel & !is.na(indices$Like)
-      subset2 <- indices$imodel==imodel
+      subset2 <- indices$imodel==imodel & indices$Yr <= endyrvec[iline]
       if(length(unique(indices$Fleet[subset2])) > 1){
         if(!is.null(indexfleets[imodel])){
           ifleet <- indexfleets[imodel]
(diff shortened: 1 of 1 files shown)


kellijohnson-NOAA:
Another something that I was reminded of when I submitting this pull request is that the local check() was generating an error. I don't have time to look into why or how to fix it this week, which is another reason why I did a pull request rather than pushing to development.

R/SSplotComparisons.R
```diff
@@ -1415,7 +1420,8 @@ SSplotComparisons <-
     }
 
     if(!add){
-      axis(1, at=yr)
+      xticks <- pretty(xlim)
```
iantaylor-NOAA: The intent of x-axis ticks at the year values was that for surveys which don't occur annually to label the actual years where the observations occurred rather than a pretty set of more evenly spaced years. However, I now believe the best practice, as used for the hake assessment, is to include dummy observations (with negative fleet) in the intervening non-survey years to provide a smoother time series of expected survey values which eliminates this effect. So thanks for this change @kellijohnson-NOAA.

PR #377, opened 2020-02-12
SS_readctl_3.30 general bug fixes

nathanvaughan-NOAA:
I have corrected a few bugs in the read control function. These were mostly due to conflicting naming conventions between read control and read data. I have changed N_areas to Nareas through out to match with the data file. This may cause back compatibility issues, but if we don't change I feel it creates confusion? If we don't think it is a good idea to change the parameter names I can modify these edits to retain back compatibility. I have also changed Ngenders to Nsexes to match with the data file structure. Finally I removed references to Nsurveys in the 3.30 script as these are no longer separated from Nfleets in 3.30.


iantaylor-NOAA:
Cleaning up the inconsistent names is definitely an improvement. Thanks @nathanvaughan-NOAA

nathanvaughan-NOAA:
I'm going to push one more fix. I accidentally changed some 3.24 things that are actually different. I'm fixing them now.

Issue #141, opened 2017-04-06
inconsistent variable names throughout r4ss


taylori:
In issue #138, @kellijohnson pointed out inconsistent variable names (Nsurv vs Nsurveys). In that case, Nsurveys is definitely more common (although this input has gone away in SS version 3.30 as surveys have become regular fleets that just don't have catch).

Unfortunately, throughout the package there are many more inconsistencies (though hopefully not within any particular function). For instance, "N_areas", "Nareas", and "nareas" all occur. I think that notation with the capital N and no space is more common, so for these variables that represent model dimensions, I would suggest doing a big find/replace to make it "Nareas", "Nfleets", and "Nsurveys". I would also change "Nfleet" to "Nfleets".

However, I don't think it's worth trying to go too crazy. In issue #77, I suggested that the package should try to follow Hadley Wickham's Style Guide: http://adv-r.had.co.nz/Style.html which says:

> Variable and function names should be lowercase. Use an underscore (_) to separate words within a name. Generally, variable names should be nouns and function names should be verbs. Strive for names that are concise and meaningful (this is not easy!).
> 
> # Good
> day_one
> day_1
> 
> # Bad
> first_day_of_the_month
> DayOne
> dayone
> djm1

I still agree with that idea and the r4ss package has tons of variable names of the "bad" variety, but I think that a large-scale clean up would be hard work and would likely break things so it would be better to just try to improve things incrementally as part of the normal process of editing functions for other, more important, reasons. 

I'm open to ideas on all this, however.

yukio-takeuchi:
I also suggest to consider consistency to what is used in control.ss_new and data_ss_new to understand the the meaning of variable easily.

For example at the top of data.ss_bew following terminologies are used in one example run

#_observed data: 
1971 #_styr
2001 #_endyr
1 #_nseas
 12 #_months/season
1 #_spawn_seas
1 #_Nfleet
2 #_Nsurveys
1 #_N_areas

When I developed SS_readctl_3.24 in particular.
I mainly  relied on those shown in control.ss_new and data.ss_new.

It might be better to collaborate the development team of SS for that purpose.

iagomosqueira:
I agree with taking it easy, but it would be good having it done at some 
point. For example, year and Yr are used as column names in different 
data.frames from SS_output (cpue and ageselex).

cgrandin:
I've changed over to using no capitals ever and dots to separate words. This is mostly because I get tired of pressing the shift key thousands of times. e.g.:

[...]

PR #380, opened 2020-02-13
fix and dev: clean up read dat and read ctl

nathanvaughan-NOAA:
I removed inconsistencies between data and control lists and moved wrapper function dependencies into version-specific data read files. This should prevent any future problems due to running the version-specific functions directly.


iantaylor-NOAA:
Looks good to me. Thanks @nathanvaughan-NOAA for getting all this done and working out the kinks to get the tests passing again.
@k-doering-NOAA could you look over the changes and merge if they seem OK to you?

k-doering-NOAA:
Hi @nathanvaughan-NOAA , I am struggling a bit with the competing goals of this PR. I think they are three fold:
1. make version specific read/write functions usable on their own. 
2. make names consistent across versions of ctl functions.
3. make names consistent across the control and data functions for the same version.

I think the first goal has been accomplished and is a great change for everyone! 
However, goals number 2 and 3 are sometimes in conflict and I am having trouble deciding how it would be best to implement them. For example, one change you suggest is to make a new parameter `Nsexes` in `SS_readctl` which should be used to specify the number of sexes in 3.30, but the old parameter `Ngenders` will be used for 3.24 models.  I understand why you thought to do this (the list component for number of sexes is `Nsexes` in the list that `SS_readdat_3.30` creates, so this makes the ctl and data code more in line), but at the same time, I think ideally we don't want users to have to use different sets of parameters for 3.24 v. 3.30 functions such that they would have to use different parameters to access them. 

The trickiness of these competing goals is probably one reason this hasn't been done in the past. I don't know if you or others (maybe @kellijohnson-NOAA or @iantaylor-NOAA ) have thoughts about how to deal with implementing goals 2 and 3?

I will be gone through Tuesday next week, so probably won't be able to contribute to this conversation until then.

iantaylor-NOAA:
Thanks for spelling out these issues @k-doering-NOAA. I would add a 4th item to the list which is even harder to fulfill in combination with the others:

4. make names consistent with data.ss_new and control.ss_new as noted in this comment https://github.com/r4ss/r4ss/issues/141#issuecomment-292305273. 

However, the suggestion there to make changes in SS for more consistency has indeed occurred, so I don't think we should be maintain any allegiance with names used in SS 3.24 output files at this stage. The biggest reason there are differences between 3.24 and 3.30 is that the 3.24 conventions were internally inconsistent so the 3.30 names represent a cleaned-up and more consistent set of variables (though obviously not perfect either). The `SS_output` function converts headers for tables read from 3.24 models to match 3.30 conventions in order to avoid having separate names for the same quantity when making plots and the like. I believe "gender" is not an accurate term for describing fish biology so I see no reason to keep "Ngenders" just because that was used in older versions of SS when "Nsexes" has completed replaced it in the SS files and should not cause too much confusion for the users. The places where inconsistencies will remain are the inclusion of terms like "Nsurveys" for quantities that exist in one version but not in the other.

Furthermore, I know that there are many SS version 3.24 models out there and we want to support them while they're still around, but if a mismatch between the terms used in the r4ss functions and the conventions used for that older version is frustrating for folks, they can either install an older version of r4ss that doesn't yet have these changes or they can upgrade their SS models to 3.30.

Finally, there's no need to fix everything in one step, and any improvement can be built on in the future. So if the tests are passing now, I have no concerns with merging stuff as it stands (after @k-doering-NOAA is back that is). I'm on vacation all next week so won't be able to contribute, but am happy to have the rest of you work on these aspects without me.

nathanvaughan-NOAA:
Sorry I didn't respond yesterday. I completely agree with all of the above points. I'm not sure either what the best approach is for correcting the conflicts between versions and file types. What do folks think about building in code that will accept both naming conventions at once (i.e. you can read in either N_areas or Nareas and whichever is changed from the default value overwrites both values, and the same for Ngenders or Nsexes) I tried to do this to some degree in the existing pull request by having Nsexes and Nareas inputs create Ngenders and N_areas values in the list object to prevent breaking users code that implements a created control file. If we go with the multiple names approach we can simply state which naming conventions are depreciated and leave them out whenever then next big 3.x update comes along.

kellijohnson-NOAA:
I might be in left field here, but I advocate for NOT worrying about matching the older versions of SS. I think consistent naming going forward is more important as a community than matching the names of older versions of SS given that the current version of SS is seen as the most robust version and the version that people should be using. Instead of over-complicating the naming scheme and having duplicates it would be better to provide more support for helping convert 3.24 models easiliy to 3.3.

k-doering-NOAA:
Thanks @kellijohnson-NOAA, @nathanvaughan-NOAA , and @nathanvaughan-NOAA for contributing to the thread - this is some helpful food for thought.

@nathanvaughan-NOAA , that is exactly what I was thinking initially, but I found that it got complicated fast to maintain back-compatibility and add warnings about deprecating function arguments eventually, mostly because we assume that people may be using the version-specific function versions directly in addition to through the non-version specific wrappers. I think it is less important that the individual list components retain the same names that they previously had, but I am think we should deprecate old function arguments over some period of time, otherwise I think it can be confusing for users.

Choices in maintaining back-compatibility vs. moving forward are problems for everyone else maintaining packages, so it might be relevant to look at what popular R packages do to deal with them. For example, the Tidyverse packages have started using [life cycle stages](https://www.tidyverse.org/lifecycle/) to indicate to users how stable certain packages or functions are, which is important given their large user base. I'm sure there are other solutions, too.

I think it might make our life easier if we took a two-staged approach to integrating the ideas in this PR: 
1) accept the changes in this PR that move code in the overall wrapper functions into the individual version specific functions;
2) then, we can open an issue and work on (in a new branch) updating function arguments and maintaining back compatibility. I am happy to contribute to this.
  - stop exporting the version specific functions and update documentation to make it clear to users that the 3.30 functions should NOT be used directly (maybe it is possible to add a warning about using them directly?). This will mean that we don't need to worry about renaming the function arguments or maintaining back compatibility, as they are not designed for users to interact with.
  - rename function arguments to as necessary and (only for the overall wrapper functions) add code to make it clear if arguments will be deprecated
I am hesitant to merge all the PR changes even thought the tests are passing, as the tests are quite simple: they just check that read/writing of a single data and single control file (3.24 and 3.30 versions) can work; in fact, they don't even check that read/writing works for a control file when the `use_datfile = FALSE` for `SS_readctl()` (although that should probably be added!). I think many of the changes here are big enough to change the way users will need  to interact with the functions and may be difficult to undo, so I don't think it is a good idea to merge everything in. 

@nathanvaughan-NOAA , what do you think about this plan? If it sounds reasonable, is there a way to cherry pick only certain commits to address step 1?

nathanvaughan-NOAA:
That all sounds good @k-doering-NOAA I think the easiest way to deal with this is for me to close this pull request, cherrypick out the changes we want to keep and then resubmit a new request. Thanks everyone for the help in planning out the best way to move forward with the changes.

PR #382, opened 2020-02-27
fix: Consolidate code into version specific read/write files

nathanvaughan-NOAA:
Moved all code out of read/write wrapper files and into version specific functions. Unified variable names between data and control read/write with previous versions to avoid any compatibility issues. Also fixed a small bug in SR_parms in readctl_3.30 that had naming errors. (Note: naming convention issues with nseas, N_areas, and Ngenders remain but can be addressed in a later commit when we decide on a best option moving forward)

### DIFF
--- a/R/SS_readctl.R
+++ b/R/SS_readctl.R
@@ -25,10 +25,12 @@
 #' also not explicitly available in control file and this information is only
 #' required if length based
 #'  maturity vector is directly supplied (Maturity option of 6), and not yet tested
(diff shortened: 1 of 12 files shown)


k-doering-NOAA:
Looks good @nathanvaughan-NOAA - thanks for all your work on this tricky issue. I added some tests to read the ctl file from the version specific functions and from the `SS_readctl()` wrapper, but when `use_datlist = FALSE`.

PR #404, opened 2020-06-26
fix(SS_output): lenntune had non-numeric values of _ for fleet 3

kellijohnson-NOAA:
I am not sure why development is currently passing because I found the error
when running check() on development.

* Copied code from agentune that changed _ to NA so type.convert can
change all values to numeric ones for mathematical operations.
* I did not remove these rows of NA as is done in agentune.
* future development should work to include a function to process the tuning
information, this new internal function could be used to process each each tuning section rather than
have code that is duplicated for the two sections.

### DIFF
--- a/R/SS_output.R
+++ b/R/SS_output.R
@@ -1927,12 +1927,14 @@ SS_output <-
 
     # Length_Comp_Fit_Summary
     if (SS_versionNumeric < 3.30) {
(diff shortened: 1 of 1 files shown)


iantaylor-NOAA:
Thanks @kellijohnson-NOAA. I don't know why it passed either. Please delete the branch of you're done with it.

PR #405, opened 2020-06-29
Develop jitter

kellijohnson-NOAA:
@iantaylor-NOAA I thought it might be cleaner to merge this into mergeSSutils and then merge that into development, but I am not sure. Let me know if you would like a different chain of events and I can submit a new pull request. This pull request adds two major functionalities to SS_RunJitter, (1) allows for users to pass two arguments to change the starter file using the same names as the list elements that are to be changed and (2) checks that the executable exists before going through and changing/moving files around.

Minor additions include the creation of a new helper function to change file names that might be useful for SS_profile as well and the use of type.convert() instead of as.numeric(). I tried to use the new formatting with more spaces but I did not change the entire file. 

Lastly, tests are now performed with a conditional if else statement depending on the availability of the executable. This test will be ran on the last folder found in the system.file() call to r4ss such that it will only test it on the newest model rather than having to build a new test every time a new simple folder is added. I don't feel like SS_RunJitter needs to be tested on all editions of SS because it should really only be ran to provide the best available science, whereas SS_output for sure has utility in performing model comparisons being able to read output from multiple versions of SS, which is why I created the test the way that I did. Please let me know if you would like the test set up in a different way.


k-doering-NOAA:
I think either will work @kellijohnson-NOAA, but it seems like it would be easier to test locally if `develop_jitter` changes work with the `mergeSSutils` changes if you merge the 2 feature branches together like you propose here. Otherwise, you would merge in the `mergeSSutils` branch to development, then submit a PR to merge in `develop_jitter` to `development`...and I think the process to get the PR changes locally is a little more complicated if you wanted to do local testing before the PR is accepted.

PR #406, opened 2020-06-29
fixes for custom reporting

k-doering-NOAA:
Small changes to get `SS_output` and `SS_plots` working with custom reporting options that will be available in SS 3.30.16. @iantaylor-NOAA, you know these functions much better than I do, so feel free to point out and make changes if the "fixes" were not done properly.

### DIFF
--- a/R/SSplotSpawnrecruit.R
+++ b/R/SSplotSpawnrecruit.R
@@ -100,10 +100,11 @@ SSplotSpawnrecruit <-
   xlab <- labels[1]
   ylab <- labels[2]
   # check if spawning output rather than spawning biomass is plotted
-  if(replist$SpawnOutputUnits=='numbers'){ # quantity from test in SS_output
+  if (is.na(replist$SpawnOutputUnits) || 
+      replist$SpawnOutputUnits == 'numbers') { # quantity from test in SS_output
     xlab <- labels[3]
   }
-  if(relative){
+  if(relative) {
     xlab <- labels[4]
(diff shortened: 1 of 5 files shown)


R/SSplotSpawnrecruit.R
```diff
@@ -100,10 +100,11 @@ SSplotSpawnrecruit <-
   xlab <- labels[1]
   ylab <- labels[2]
   # check if spawning output rather than spawning biomass is plotted
-  if(replist$SpawnOutputUnits=='numbers'){ # quantity from test in SS_output
+  if (is.na(replist$SpawnOutputUnits) ||
```
iantaylor-NOAA, line 103: Defaulting to numbers rather than biomass when units are unavailable here differs from the approach I took in `SSplotTimeseries` which defaulted to biomass:
https://github.com/r4ss/r4ss/blob/development/R/SSplotTimeseries.R#L156-L157.
However, although biomass may be the more common units, the term "spawning output" can be inclusive of biomass as opposed to "spawning biomass" which does not work if the output is in some other unit. I will change `SSplotTimeseries` to match this better approach.

iantaylor-NOAA:
Thanks @k-doering-NOAA!

R/SSplotSpawnrecruit.R
k-doering-NOAA: That was my thinking, that if the units couldn't be found, the more general "spawning output" term could be used.

PR #407, opened 2020-07-01
Merge ssutils

kellijohnson-NOAA:
@iantaylor-NOAA can we merge this into development and delete the branch or are there still things that we need to do here? I am worried that with changes to the namespace file that the merge could get messy if we wait too long.


iantaylor-NOAA:
Done. Thanks for the reminder @kellijohnson-NOAA

PR #410, opened 2020-07-22
Fix jitter

kellijohnson-NOAA:
Allow for new custom reporting file in SS_RunJitter by using summary.sso for those models that produce it. To provide backwards compatibility, even though users should upgrade, SS_output is available to users when summary.sso is not available. Also, tried to clean up some functions by moving code that may apply to other functions external to SS_RunJitter().


iantaylor-NOAA:
Thanks @kellijohnson-NOAA for these improvements. It will be good to use `check_model()` elsewhere, such as these spots: https://github.com/r4ss/r4ss/search?q=version%24os&type=Code

PR #416, opened 2020-08-11
fix: movement in ss_output

k-doering-NOAA:
headers were being created from the first row when they are already being pulled in matchfun2. Thanks to Dan Fu for reporting this bug (https://vlab.ncep.noaa.gov/web/stock-synthesis/public-forums/-/message_boards/view_message/11095850)

@iantaylor-NOAA could you make sure this change looks ok and merge in if so?


iantaylor-NOAA:
Thanks @k-doering-NOAA !

PR #422, opened 2020-09-10
Functions for the calculation of Empirical selectivity diagnostics.

roliveros-ramos:
The function SS_output has been modified to calculate the empirical selectivity when reading SS model outputs.


iantaylor-NOAA:
Thanks @roliveros-ramos for putting this together. I've merged into a new `empirical_selectivity` branch within the `r4ss` organization to allow any refinements to be separated from a an ongoing bug fix I've got in the works in `development`. The Travis-CI fail is simply from the testthat check of `SS_output()` looking `$inputs` as the final element of the list which is returned and seeing `empirical_selectivity` instead, but easy for me to fix.

PR #437, opened 2020-11-10
Update SS_output.R

AdrianHordyk:
replist$exploitation wasn't being imported correctly prior to this change. (SS V3.30.16.00)

### DIFF
--- a/R/SS_output.R
+++ b/R/SS_output.R
@@ -110,6 +110,9 @@ SS_output <-
       })
     }
 
+
+
+
     matchfun2 <- function(string1,
                           adjust1,
                           string2 = NULL,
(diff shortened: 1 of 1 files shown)


iantaylor-NOAA:
Thanks for catching this @AdrianHordyk.

I have attempted to resolve the problem in an alternative way in commit d760bdc to the `development` branch. Changes to to the `Report.sso` file and the `SS_output()` function earlier this year were driven by a new flexibility within SS to choose which elements get reported, and I've tried to now make the parsing of each table depend only on the blank line at the end, rather than the keyword at the start of the following section (as discussed in issue #9). However, there are hash marks that in some cases need to be ignored and in other cases have meaning as separators, causing bugs like this one.

Could you test the `development` branch and see if this fix it works for your model?

AdrianHordyk:
In my case is was an extra hash mark causing the problem - appears to be working fine now in `development` branch. 
Thanks

iantaylor-NOAA:
Good to hear @AdrianHordyk, thanks for checking.

Issue #438, opened 2020-11-10
processing initial F parameters in SS_readctl_3.30()


iantaylor-NOAA:
The `stop()` on these lines doesn't seem to work well when running SS_readctl() using default inputs because `N_rows_equil_catch = 0` so it gets skipped:
https://github.com/r4ss/r4ss/blob/development/R/SS_readctl_3.30.R#L778-L785
This leads to the error `Error in `.rowNamesDF<-`(x, value = value) : invalid 'row.names' length` when the Q_setup table gets read incorrectly.

@k-doering-NOAA do you have ideas on how best to at least improve the error/warning?

k-doering-NOAA:
This is a tricky one.  Maybe  this could be a warning (as opposed to stop) regardless of the value of `N_rows_equil_catch`, if `use_datlist = FALSE`? This would mean you may get the same strange error, but a potentially more helpful warning message.

I wonder if `N_rows_equil_catch = NULL` should be a valid option and the default. In this case, SS_readctl_3.30 will attempt to figure out the equilibrium catch inputs automatically if `use_datlist = FALSE`:

1. use keywords in the original file's comments to find the equilibrium catch vals
2. if any are found, see if they match with the expected next inputs
3. If they don't match, the SS_readctl_3.30 errors; if they do, then they can be read in.

Another route we could go is require a datafile always when using `SS_readctl` ( or at least `SS_readctl_3.30`) - however, I could see how it may be useful to be able to read in a ctl file without a data file by specifying the necessary values.

iantaylor-NOAA:
Thanks for the input on this, @k-doering-NOAA.
I like `N_rows_equil_catch = NULL` and an attempt to figure out how many initial F parameters there are. I'm no longer in touch with what approaches are being used in `SS_readctl_3.30`, but a few indications (none perfectly reliable) that the model contains initial F parameters are 

- a line like `#_initial_F_parms; count = 2`
- 7 numerical values on at least one row between the `F_Method` input and `Q_setup`.
- numerical values followed by a comment (parameter label) that begins "InitF"

I like the idea of a test using something like one of the above, even if imperfect, before providing a warning and/or attempting to parse the lines appropriately. If a warning was displayed every time `use_datlist = FALSE` I think it would get routinely ignored, even for those situations where it was useful.

k-doering-NOAA:
[...]

PR #440, opened 2020-11-16
Eliminate some 3.24 list elements from SS_readctl_3.30, address #438

k-doering-NOAA:
In `SS_readctl` with a 3.30 model, some list elements created were done for back compatibility with 3.24. However, this can be confusing for users, as it is not clear which elements will be written using `SS_writectl` and which are only available for back compatibility purposes. Thus, some of these elements have been removed, including `N_cpue` and `N_CPUE_obs`, `Variance_adjustments` (`Variance_adjustment_list` remains, as it is the 3.30 formatted version), and `Q_setup` (`Q_options` remains). 

For those who want to convert the 3.30 variance adjustments and Q_options to the 3.24 format, this is now possible to do using the new functions `translate_3.30_to_3.24_var_adjust` and `translate_3.30_to_3.24_Q_setup`  (I'm guessing @nklaer may want to know about this, but I expect this functionality will not be necessary for most users). These new functions take the 3.30 objects as input.

Lastly, this address #438 . Full details are available on the issue, but basically this changes the default for `N_rows_equil_catch` in `SS_readctl` to be `NULL` when a data file is not read in and used when reading the control file. The function will attempt to use comments on the line `#_initial_F_parms; count = 0` in the file to determine how many equilibirum values should be read. If that number on that line is incorrect, or if the comment is missing, an error will be generated. Previously, all calls with `use_datlist = FALSE` and any equilibrium catch would exit on error, so this expands the number of models that can be read when using `use_datlist = FALSE`.


iantaylor-NOAA:
Looks good to me, @k-doering-NOAA.
Should Roxygen comments for the new `translate_3.30_to_3.24_Q_setup()` function include `@export` so it appears in `NAMESPACE`? I think it would still be available as `r4ss:::translate_3.30_to_3.24_Q_setup()` but that could be confusing for folks. Then again, it's hard to imagine very many folks using this so maybe it's better to keep the visibility low.

k-doering-NOAA:
Yes, I wasn't sure if these functions should be exported or not? I settled on not exporting to keep their visibility low, but I'm not sure that is the right choice. It would be useful to hear what others think.

Also, they could be renamed to something more concise - I just couldn't think of anything concise but descriptive.

iantaylor-NOAA:
I'm happy keeping things as you've left them. Thanks again for all this cleanup.

PR #441, opened 2020-11-17
Allow relative path to exe in run_SS_models, fix to check in SS_tune_comps

k-doering-NOAA:
This PR partially addresses #430 by allowing relative paths to the executable in `run_SS_models`. It also adds some additional checks on where the exe is located. Note, however, it is not perfect; The biggest issue I'm not sure how to fix is that if you wanted to use an exe called "ss.exe" in each model folder specified in dirvec, but ss.exe was also in the user's PATH (environmental variables) , the PATH version would be used by default even if `exe_in_path = FALSE`.  I have added a warning when this happens, but there may be a cleaner solution (for example, maybe we don't really need the `exe_in_path` argument at all?

This also includes a fix to a check on the `fleets` argument in `SS_tune_comps` that wasn't working properly.

### DIFF
--- a/R/SS_tune_comps.R
+++ b/R/SS_tune_comps.R
@@ -127,10 +127,15 @@ SS_tune_comps <- function(replist = NULL, fleets = "all",
     verbose = FALSE
   )
   if (fleets[1] == "all") {
-    fleets <- 1:dat$Nfleets
+    fleets <- seq_len(dat$Nfleets)
   } else {
-    if (length(intersect(dat$Nfleets, 1:dat$Nfleets)) != length(fleets)) {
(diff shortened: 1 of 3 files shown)


iantaylor-NOAA:
This all looks great, @k-doering-NOAA. 
Regarding the challenge of dealing with the situation where the SS executable in the path has the same name as the one in their model directory, they will have this challenge on the command line as well. Renaming the exe in the path to a version-specific filename (as I think you typically do) might be the best solution to this.

I'll merge after the Travis checks have completed, but can't wait for that today, so you're welcome to merge yourself if you're online when you see that happen.

Issue #442, opened 2020-11-19
partial matching of using $ for list elements can cause bugs


k-doering-NOAA:
Thanks to @shcaba for reporting a recent example of this. (fixed in 96dfa2e)

It is often easier/faster to use `$` instead of `[[` to reference elements of lists in r4ss, because R studio will help you complete the expression. For example, `dat$CPUE` as opposed to `dat[["CPUE"]]`. These are generally equivalent, except that `$` allows partial matching while `[[` does not.

This can cause issues when list components can be NULL (thus, not produced), but partially matched with other existing list components. for example, `dat$CPUE` will match `dat$CPUEinfo` if the CPUE does not exist (i.e., is NULL). However, dat[["CPUE"]] will correctly return NULL.

This pops up occasionally creating bugs, but not that often. Some possibilities on how to deal with this:

1. Status quo: push fixes when these bugs come up.
2. Write a script that can parse the SS source code and replace all instances of $ with the `[[` formatting? Is there times when `$` does need to be used and changing would mess it up?
3. assign list components with NA, rather than conditionally creating them. I don't think there would be an automated way to do this, however, and it seems like it could be more work?

Anyone have thoughts on this?

iantaylor-NOAA:
Option 2 sounds good to me although it would probably take a lot of manual checking to make sure the changes all made sense.

Regarding Option 3, I think that missing objects were NA long ago and I shifted to using NULL instead because 
`if (is.null(some.data.frame))` 
seemed cleaner than 
`if (!is.na(some.data.frame[[1]][1]))`
My sense is that it wouldn't be worth the effort to make the switch back to NA but if there would be other benefits beyond the partial matching problem, I could be convinced otherwise.

roliveros-ramos:
There is another possibility for option 3, setting the elements to NULL:

```
dat = list(CPUEinfo=1:10)
dat$CPUE
```
[1]  1  2  3  4  5  6  7  8  9 10

```
dat = list(CPUEinfo=1:10, CPUE=NULL)
dat$CPUE
NULL
```
After a list has been created, names can be checked again a master names list and missing names added with NULL as value. That would eliminate the potential problems with partial matching. Or, start creating a list filled with names set to NULL.

k-doering-NOAA:
Thanks @roliveros-ramos !  I've thought about that solution before, but I think sometimes when you set list values to NULL their names disappear from the list entirely? (Maybe you know more about the nuances of this?)

For example, if I run:
```r
[...]

Issue #399, opened 2020-05-18
use styler package to apply automatic formatting to all r4ss functions


iantaylor-NOAA:
r4ss is full of legacy code written long before any of the contributors were thinking of applying a consistent style. The `styler` package (https://styler.r-lib.org/) includes functions to automatically apply elements of the tidyverse style guide (https://style.tidyverse.org/) by adding things like spaces around operators and line breaks to long lines.

I think that using `styler` to clean things up will make the code easier to read, easier to debug, and easier to contribute to. 

However, anyone who has a fork or branch of r4ss could face lots of merge conflicts when trying to merge any changes to the pre-styled code with the post-styled code. Thus, this issue is both a call for input on the use of styler from anyone with opinions about it and a suggestion to keep your forks/branches up to date.

I expect the remaining steps will look something like this:
- [x] delete merged branches in r4ss repository
- [x] clean up unmerged branches/forks as needed
- [ ] test out different options within `styler` to find the right fit (thorough clean-up but not too different from status-quo)
- [x] apply `styler`
- [ ] attempt to rebase unmerged branches or forks as needed

iantaylor-NOAA:
Changes to `SS_output()` associated with PR #402 were made after first running the default styler changes via `styler::style_file('SS_output.R')` (in commit c4babc59700c504e04b48bbb4b7e1660908ab2a2). I found the reformatted code easy to work with although applying the same styler function again (see commit c10d61e) after my manual changes revealed that modifying my coding style will take some work as I routinely continued to use `if(...){` to which styler adds spaces: `if (...) {`. Also, the default indentation applied by my editor (Emacs with ESS) didn't match that used by styler when function inputs span multiple lines. Nevertheless, I'm convinced that routinely applying styler to all r4ss functions to apply a consistent coding style is a good idea and will proceed to do so when I have time to confirm that it doesn't break anything.

PR #447, opened 2020-12-04
Issues #442 and #399

k-doering-NOAA:
Addresses #442 and #399 (removing dollar sign subsetting, which allows for partial matching, and use styler to get consistent styling for the r4ss source code).


iantaylor-NOAA:
Thanks @k-doering-NOAA. It will be great to have this more robust and cleaned-up code going forward.
Let's confirm tomorrow that the checks have passed and then merge this pull request.

k-doering-NOAA:
Sounds good! One thing I should mention is if the dollar sign was intentionally used to allow partial matching, then the "cleanup" would cause the code to break. I think this situation should be pretty rare, though...

k-doering-NOAA:
@iantaylor-NOAA , I think this is ready to be merged. commit 43b5a53 was pushed to fix code that worked with partial matching before (I think only needed by models with tagging data) that was converted to not work with partial matching. I think partial matching was used just for convince?

We might catch a few more bugs like this where partial matching was used, but I think these issues will be few and far between.

iantaylor-NOAA:
Thanks for all this work, @k-doering-NOAA. As discussed offline, I'm merging after one more minor change and we'll wait for the tests to re-run on the merged development branch.

Issue #448, opened 2020-12-08
set up github action workflow to run testthat tests that require the ss exe


k-doering-NOAA:
Set up a github action that can run the testthat tests that require the ss exe. This would help ensure that the tests for these functions are tested more regularly. 

The steps generally are:
1. Get an ss exe from the ss website
2. Download the current (dev branch) version of r4ss, devtools, and other dependencies
3. Run the testthat tests that require the exe. This may be just a few scripts, or if tests requiring the exe are mixed throughout, we could just run all the tests, since the ones that require an exe are likely to be the ones that take the most time, while the ones that don't should run quickly.

This will likely take some trial and error, but should be possible to set up.

k-doering-NOAA:
I should add, this was @iantaylor-NOAA 's great idea!

PR #457, opened 2020-12-23
Develop new gh action to run tests w/ SS executable

k-doering-NOAA:
Addresses issue #448 . Sets up a new action that run Testthat tests in test-tune_comps.R and test-runs.R ,which require the SS executable to run. Some notes: 
1. The build is currently failing due to a failing test; it seems that SS_doRetro is not working for the Simple 3.30.01 model; I have not yet investigated why (other than the retro models did not run), but I figure this can be addressed after merging in this branch.
2. For some reason, devtools::test() couldn't find the kableExtra pkg, even though kableExtra was defintely being installed in the R library for the run. For now, there is a hack in the script to move kableExtra from "depends" to "suggests" in the DESCRIPTION (only for the gh action workflow, nothing is changed in the r4ss repo). This solves the problem temporarily, although hopefully it can be solved in a more elegant way in the future.
3. Putting together this run was helpful in identifying some differences between what works on windows and what works on unix systems; For example, using someting like  `is.exe <- file.info(dir(dir.old, full.names = TRUE))$exe` does not work on linux/osx, b/c there is no $exe list component. I'm not sure what the os independent solution to this is, but we will need other ways to check for the presence of the stock synthesis executable in future code.

### DIFF
--- a/.github/workflows/R-CMD-check.yml
+++ b/.github/workflows/R-CMD-check.yml
@@ -1,6 +1,6 @@
 # For help debugging build failures open an issue on the RStudio community with the 'github-actions' tag.
 # https://community.rstudio.com/new-topic?category=Package%20development&tags=github-actions
-on: [push, pull_request]
+on: [workflow_dispatch]
 
 name: R-CMD-check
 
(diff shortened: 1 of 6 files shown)


k-doering-NOAA:
I also should add, I've just left this job as manually triggered for now, rather than on push/PR or on a schedule. I'm not sure how often we should run these tests. Maybe once the code has been fixed so the runs pass it should be run more often than manually?

iantaylor-NOAA:
This is really useful, thank you @k-doering-NOAA!
The codecov metric doesn't reflect the huge improvement because it just keeps track of which tests exist, not which ones actually get run.

PR #463, opened 2021-01-12
Fix SS_output so that tests work for MSE

cgrandin:
* Weird problem where "_" > 0 is TRUE
* Trailing space removed due to script automatically doing it (not related)


iantaylor-NOAA:
Looks like @kellijohnson-NOAA also fixed this in the main code back in June: https://github.com/r4ss/r4ss/commit/816f3368622475a956dbd31e5a1fd2ea4753ccf0

The `hake2020` commits that are ahead of the `development` branch are pretty small so it may be worth trying to either switch over to `development` or `master` at some point in case that doesn't break anything.

k-doering-NOAA:
In case it is helpful, this shows all the diffs between development and hake2020 branches: https://github.com/r4ss/r4ss/compare/development...hake2020

PR #469, opened 2021-01-21
fix(SS_tune_comps): MI_mult for agecomps is wrong in SS 3.30.12-16

kellijohnson-NOAA:
QUESTIONS: 
* can someone check my proposed math in this commit to
ensure that I got the tuning correct?
* also, now that I created the pull request I remembered that
I only tested this with age comps; can someone check that I didn't
break things for length comps?

recommended variance adjustment is wrong b/c it was set to 1.0.
@iantaylor-noaa knows and is fixing it in SS but r4ss will still
need to work with this output for 12-16 models.

Updated documentation based on ability to use markdown in the file,
which makes things easier to read, moved some sections.

Noticed during the 2021 pacific hake stock assessment

### DIFF
--- a/R/SS_tune_comps.R
+++ b/R/SS_tune_comps.R
@@ -1,44 +1,75 @@
-#' Calculate new tunings and (re)run models for length and age compositions
+#' Calculate new tunings for length and age compositions and (re)run models
 #'
 #' Creates a table of values that can be copied into the SS control file
 #' for SS 3.30 models to adjust the input sample sizes for length and age
-#' compositions based on either the Francis or McAllister-Ianelli tuning.
-#' Optionally, this function can also automatically add the tunings and rerun
-#' the model for the number of iterations desired by the user.
+#' compositions based on either the Francis or McAllister-Ianelli tuning or
+#' adds the Dirichlet-Multinomial parameters to the necessary files to
+#' tune the model using an integrated method.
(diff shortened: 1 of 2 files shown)


k-doering-NOAA:
The tests in https://github.com/r4ss/r4ss/blob/development/tests/testthat/test-tune_comps.R are run with the gh actions "run-tests-with-SS" job. It uses a length comps model. Kelli, are you concerned that the values aren't right for len comps, or just want to make sure it runs? I think these tests show you didn't introduce code that causes the function not to run. I can look in more detail at len comp weights produced if you think it is necessary.

kellijohnson-NOAA:
I just wanted to provide the disclaimer that I might not have thought through all scenarios given I was only working with age data. Though, I just checked that with length data (copper rockfish 2021; thanks @chantelwetzel-noaa ) it gives you the same output as the Recommended_var_adj. I will note @k-doering-NOAA that your tests altered me to an earlier dumb commit that I got rid of with -f early this morning. So thank you for your providing those.

kellijohnson-NOAA:
I pushed a new commit because using -f would get rid of my mistake and the context of this discussion on markdown would seem weird. So, I will leave my mistake documented for eternity :(

iantaylor-NOAA:
Checks have passed, seems adequately tested on copper example. Thanks @kellijohnson-NOAA.

PR #472, opened 2021-01-25
Killoutput

kellijohnson-NOAA:
Working with ss3sim that uses SS_tune_comps to reweight data behind the scenes, I am trying to limit the output that is printed to the screen if verbose = FALSE. Which led to two things
1. SS_tune_comps() reads in report files with SS_output multiple times without setting covar = FALSE or TRUE. This is okay, but I put in a check to see if you for sure are not running with the hessian then do not try to have covar = TRUE. This essentially doesn't change anything. I did note that if you set covar = FALSE and you do have the hessian, that the results of Francis low and Francis hi will be different. Should we not be reporting these values if there is no hessian? Doesn't matter for weighting though.
2. show.output.on.console = FALSE stops admb output from being printed to the screen. I know that there are weird combinations here of arguments inside of system that either turn printing off or don't and I don't claim to know all of them, but this works for what I need :)

### DIFF
--- a/R/SS_readdat_3.30.R
+++ b/R/SS_readdat_3.30.R
@@ -69,7 +69,7 @@ SS_readdat_3.30 <-
       stop("Error - There was no EOF marker (999) in the data file.")
     }
     if (is.null(section)) {
-      if (Nsections > 1) {
+      if (Nsections > 1 & verbose) {
         message(
           "The supplied data file has ", Nsections,
           ifelse(Nsections == 1, " section. ", " sections. "),
--- a/R/SS_tune_comps.R
+++ b/R/SS_tune_comps.R
@@ -283,13 +283,16 @@ SS_tune_comps <- function(replist = NULL, fleets = "all",
         skipfinished = FALSE, verbose = verbose,
         exe_in_path = exe_in_path, ...
       )
-      replist <- SS_output(
-        dir = dir, verbose = FALSE, printstats = FALSE,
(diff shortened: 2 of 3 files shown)


iantaylor-NOAA:
Thanks for these improvements @kellijohnson-NOAA.

Regarding the Francis tuning differences in 95% intervals, I see that there is a random element in the calculation through the use of `sample()` on this line:
https://github.com/r4ss/r4ss/blob/2e5f0d5349c4e269fc6018be81f95f137360ff6c/R/SSMethod.TA1.8.R#L350
Thus, running for instance the function on the same model twice will produce slightly different 95% intervals as shown below. I suspect that's the source of the difference you're seeing when running this on objects created with `covar = TRUE` vs `covar = FALSE`. If you think it's something more significant, let me know.
```
> SSMethod.TA1.8(model, type='len', fleet=1)
[1] "Francis Weights - len: FISHERY: 0.2735 (0.1749-0.5563)"
        w        lo        hi 
0.2735316 0.1748812 0.5562604 
> SSMethod.TA1.8(model, type='len', fleet=1)
[1] "Francis Weights - len: FISHERY: 0.2735 (0.1698-0.6)"
        w        lo        hi 
0.2735316 0.1697550 0.6000100 
```

kellijohnson-NOAA:
Oh I see, thanks for digging into it and merging the pull request. I will delete the branch now.

PR #474, opened 2021-01-26
Fix subplot=11 of SSplotTimeSeries and improve label of time block parameters of SS_readctl_3.30 

yukio-takeuchi:
Subplot=11 (Total recruitment time series) in SSplotTimeSeries has a bug if there is multiple areas .  This commit fix this.

Also SS_readctl_3.30 is improved on the label(rownames) of time vary parameters for time block. Now it reflects block function (column 14 of parameter).  Previously "add" is always used regardless of block function used.


iantaylor-NOAA:
Looks good. Thanks for your help @yukio-takeuchi!

PR #476, opened 2021-01-26
fix(SSsummarize): DM parameters have a new name (DM_theta);

kellijohnson-NOAA:
new row in pars, parSD, and parphases

* write a helper function copy.dm, where the DM parameters that were named
EffN_mult are copied into the rows for the DM_theta parameter which will be in
newer models; nothing is changed in the EffN_mult line, this is to encourage people to
use the new naming rather than leaving their code as is with EffN.
* name parphases with model names as is done for the other "par" data frames;
I looked back in the code and this line had not been touched other than line endings
for over 10 years and I am guessing it is just that no one uses this table that
has led to the names staying V1 V2 ... instead of model1 model2.
* make a few if(verbose) checks to limit output to the screen.

pushing to branch to run automated tests prior to merging in, but I don't see
any problems in merging this into development

### DIFF
--- a/R/SSsummarize.R
+++ b/R/SSsummarize.R
@@ -95,7 +95,9 @@ SSsummarize <- function(biglist,
   for (imodel in 1:n) {
     stats <- biglist[[imodel]]
     listname <- names(biglist)[imodel]
-    message("imodel=", imodel, "/", n)
+    if (verbose) {
+      message("imodel=", imodel, "/", n)
+    }
 
     # gradient
     maxgrad <- c(maxgrad, stats[["maximum_gradient_component"]])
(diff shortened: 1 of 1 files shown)


iantaylor-NOAA:
Thanks for these changes @kellijohnson-NOAA.
It occurs to me that `SSsummarize()` doesn't have a way to know if the D-M parameters, regardless of old vs. new label, are associated with the same fleet/data type in each model. So if the user has changed those assignments, it's on them to interpret the parameters appropriately.

kellijohnson-NOAA:
Great thought @iantaylor-NOAA ... I think that it is fine to assume that users need to know that we are only matching parameter names here and not fleets because the parameter name itself doesn't indicate the fleet where the selectivity parameters do. I will add a if (verbose) message() to copy.dm()

PR #487, opened 2021-02-12
Add error check for need_catch_mult in fleet_info

yukio-takeuchi:
4th column of fleet_info indicates if catch multiplier is to be used by setting this column to 1.  Catch multiplier can only be used for catch fleets. However SS_readdat_3.30 does not check if 4th column is 1 for fleets other than catch fleets. In the case of SS, it stop reading datfile if 4th column of fleet_info is 1 for non-catch fleets. This PR add similar error check for 4th column of fleet_info.


k-doering-NOAA:
Good idea, @yukio-takeuchi! The SS_read functions could definitely use more checks like this. I slightly modified to the message so it will print better if there are multiple fleets using catch multiplier but shouldn't (probably a rare situation), but looks great- I will merge in once checks pass. Thanks for adding this!

Issue #493, opened 2021-03-10
Remove lines of 0 comps when reading in data files


k-doering-NOAA:
Thanks @iantaylor-NOAA for suggesting the following:

The r4ss::SS_writedat() function could also be improved to check for 0 [total fish in a line of comps] and skip writing those vectors with a warning.

If there are 0s in length or age comps, SS exits on error, so it makes sense for the r4ss to helpfully remove these lines.

PR #494, opened 2021-03-15
feat, issue #493: remove comp data lines with zeros

k-doering-NOAA:
See #493 . @iantaylor-NOAA , could you make sure this looks right when you have a chance?

### DIFF
--- a/R/SS_readdat_3.30.R
+++ b/R/SS_readdat_3.30.R
@@ -413,6 +413,17 @@ SS_readdat_3.30 <-
             }
           )
       }
+      # remove any 0 values in the lencomp, because SS will exit on error
+      # if there are any lines of all zero age comps
+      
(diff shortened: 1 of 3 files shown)


iantaylor-NOAA:
Thanks for working on this @k-doering-NOAA.
Thinking this through just now, I think it would be better to just have a warning in `SS_readdat_3.30()` and save the actual filtering for `SS_writedat_3.30()` (also with a warning). The zero-value rows might be the result of some data processing issue that users may want to fix at some other point in their workflow, and having them available in the list created by the r4ss function would help them identify where the problem lies.

k-doering-NOAA:
Thanks @iantaylor-NOAA , good suggestion. I've modified this so that lines of 0 agecomps or length comps are only removed in SS_writedat rather than in SS_readdat.

iantaylor-NOAA:
Thanks @k-doering-NOAA.
Nice work adding the automated test of this new warning/filtering setup!

Issue #324, opened 2019-07-11
Negative CPUE vals


kellijohnson-NOAA:
https://github.com/r4ss/r4ss/blob/5f2dc612a5aa15e0d13f3548ae354bc9a5300db8/R/SSplotIndices.R#L205

In sablefish, we are using a mean centered index of recruitment. This index is not being plotted correctly because some of the values are negative and are being removed here with the CPUE > 0 call.

kellijohnson-NOAA:
Maia is fixing it now, but I wanted to document it to see if there is some reason why we need to remove negative index values?

iantaylor-NOAA:
The SS User Manual says "Previously the code for not using observations was to enter the observation itself as a negative value. However, that old approach prevented use of a Z-score environmental index as a 'survey'." This text was also in the 3.24 manual, so I'm not sure how long ago "previously" was. But since the current approach (using negative fleet ID to exclude values) has been in place since at least 3.24, I think it's safe to remove all occurrences of `cpue$Obs > 0` in SSplotIndices. Thanks @mkapur for working on this.

For future reference, issue #64 is related (an option to show index values with negative fleet number might be useful depending on the reason for exclusion).

kellijohnson-NOAA:
Thanks @iantaylor-NOAA for the comments and sorry I didn't take the time to find the other issue. @mkapur found an additional prob with negative values because of log transformations. She is working on how to best deal with this.

iantaylor-NOAA:
It looks like `survey_error` is a column in a table at the end of `DEFINITIONS` in Report.sso. This gets returned as `$survey_error` in the list produced by `SS_output()`.

`SSplotIndices()` should be changed so that any fleet with `$survey_error = -1` (normal) or `$survey_error > 0` (T-distribution) do not have the log-scale plots produced. The use of `qlnorm` should also be modified in these cases to use `qnorm` or `qt` as needed.

@mkapur, let me know if you want me to tackle any of this--I'm not the one in the review meeting.

mkapur:
Hey Ian
Only if you have the time/bandwidth that would be great — I’m about to
begin a long commute and wasn’t going to touch this for another 2 hours. So
your familiarity with the function may make things happen faster. No matter
what I’ll take a look when I’m back at my computer later.
Thanks!

iantaylor-NOAA:
[...]

PR #495, opened 2021-03-16
More for #324 sablefish environmental index in figures.

kellijohnson-NOAA:
@taylori this really simple change is a pull request for you because you were the one that worked with #326 and #324. Sablefish models are still running into errors and I tracked the latest one down to this where `index_resids.fn` doesn't play nice with the normal distribution.

@mkapur this fix should fix the error you were having with building the pdf.

### DIFF
--- a/R/SSplotIndices.R
+++ b/R/SSplotIndices.R
@@ -772,7 +772,7 @@ SSplotIndices <-
 
         # residual/deviation plots
         if (plot) {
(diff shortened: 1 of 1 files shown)


iantaylor-NOAA (approved):

iantaylor-NOAA:
Thanks @kellijohnson-NOAA.

PR #502, opened 2021-03-24
Jenkins r4ss testing script 

k-doering-NOAA:
@iantaylor-NOAA , this seems to be working consistently, so I think it would be good to merge into development? Thanks for figuring out this fix!


iantaylor-NOAA:
Thanks for remembering to sort this out, @k-doering-NOAA.

PR #506, opened 2021-03-26
allow SS_readctl_3.30 to read tv trends, small fixes

k-doering-NOAA:
- SS_readctl_3.30 can now read time varying trends. Before, would exit on error with a confusing warning.
- fix: SS_readctl_3.30 correctly reads the number of parameters for env link option 4, which requires 2 short parameter lines, not 1 like the other options.
- feat: adds a warning to SS_readdat_3.30 if the data file does not have consecutively numbered Dirichlet Multinomial parameters starting from 0. SS will exit on error while reading input files, but warning instead of error allows  users to fix this within the list read in using SS_readdat.
- fix: A check that there are no lines of all 0 length comps errors out when there are no length comp lines. Adds in a fix so this check is skipped if there are no lines.


iantaylor-NOAA (approved):

iantaylor-NOAA:
Thanks @k-doering-NOAA

PR #513, opened 2021-04-07
Updates to SS read and write control and data file functions

k-doering-NOAA:
These are as described in issue #505 to have defaults that are more often used.

In SS_readctl (and the functions it wraps, SS_readctl_3.24() and SS_readctl_3.30()):
- `use_datlist = TRUE`  and `datlist = data.ss_new` are now the default values in `SS_readctl()`
- Default inputs used with `use_datlist = FALSE` are now set to `NULL`
- `Ngenders` has been soft deprecated and replaced with `Nsexes`
-  `verbose = FALSE` is now default
- some unneeded arguments were soft deprecated. (`echoall`, `version` in the lower level functions)

In `SS_readdat()` and `SS_writedat()` (breaking change!):
- The `Ngenders` list component of the list created by `SS_readdat()` has been duplicated as a list component named `Nsexes`. Users planning to modify the number of sexes in their data file should modify `Nsexes`, as eventually the `Ngenders` list component will be removed. `SS_writedat` writes out the value of `Nsexes`, but warns if `Nsexes` and `Ngenders` are not equal.


iantaylor-NOAA:
Thanks for making all these changes @k-doering-NOAA. It ended up being more involved that my initial suggestion to I appreciate the help.

Issue #553, opened 2021-09-27
updates to read and write functions for 3.30.18 release


k-doering-NOAA:
- [x] Read and write ctl and data for new predator fleet option
- [x] Read and write forecast for output for F at Blimit option
- [x] Read and write ctl for new F type 4 option.

PR #554, opened 2021-09-28
Read/write function changesahead of Stock Synthesis 3.30.18 release

k-doering-NOAA:
Addresses #553  to allow reading and writing of new Stock Synthesis features.


iantaylor-NOAA:
Changes look fine to me.
There were some failed checks with the Ubuntu devel platform but those seem to be related to a package dependency issue with `rcmdcheck` (see https://github.com/r4ss/r4ss/pull/554/checks?check_run_id=3737182829#step:12:14280), which seems to be supported by RStudio so likely just a short-term problem that won't impact any r4ss users.

PR #568, opened 2021-11-04
Update scripts in SSmohnsrho

brianlangseth-NOAA:
Ive made the following changes to the calculations:
1. Previous Mohn's rho calculation for the original (option 1) and Woods-Hole (option 2) types were off by one year. The value for the first peel was calculated as the relative difference at the terminal year of the base model, and the value for the second peel was calculated as the relative difference at the terminal year of the first peel, and so on. Thus, Mohn's rho for a 5 year retrospective from from a base model ending in 2020 was calculated from the differences at 2020-2016. The differences should be calculated at 2019-2015. 
My fix was to add "+1" to the index of the endyrvec when calculating the original and Woods-Hole Mohn's rho. 

2. The Woods-Hole rho calculations only calculated the differences among years where data was removed. Based on an [ICES Legault (2009) report](https://archive.nefmc.org/tech/council_mtg_docs/Sept%202009/Herring/Doc%209_Retro%20Working%20Group%20Report.pdf) and an [AFSC Hanselman et al (2013) report](https://www.researchgate.net/publication/349830067 ), the differences should be calculated from all years in each peel (going back to the start year of the model). Thus a 5 year retrospective run from a 2020 model calculated differences only back to 2015, but should calculate back to the start year of the model.
My fix was to set the default for the startyr to be the starting year from the reference model, and clarified this in the function description.

3. The fix to the original mohn's rho calculation now gives the same value as the code script for the AFSC - Hurtado-Ferro type. To streamline the code and remove duplication, I removed the code block for the AFSC - Hurtado-Ferro, and moved the heading section to the end of the function, where the original Mohn's rho is divided by the number of peels to obtain the AFSC - Hurtado-Ferro value.

4. I removed initialization of the HmohnSSB.all variables because these weren't used anywhere. 

I also added some additional text describing the different types of values the function calculates to make more clear how type 1, 2, and 3 differ. In the process, I added the reference to Mohn 1999.

### DIFF
--- a/R/SSmohnsrho.R
+++ b/R/SSmohnsrho.R
@@ -1,10 +1,12 @@
 ##' Calculate Mohn's Rho values for select quantities
 ##'
 ##' Function calculates:
-##' (1) a rho value for the ending year for each retrospective relative to the reference model,
-##' (2) a "Wood's Hole Mohn's Rho", and
-##' (3) an "Alaska Fisheries Science Center and Hurtado-Ferro et al. (2015) Mohn's rho"
-##' (2) and (3) are based on all years between the reference and the retrospective run.
+##' (1) a rho value for the ending year for each retrospective relative to the reference model
+##' as in Mohn (1999),
+##' (2) a "Wood's Hole Mohn's Rho", which is a rho value averaged across all years for each
+##' retrospective relative to the reference model, and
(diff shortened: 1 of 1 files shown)


Huihualee-NOAA:
Hello Brian

I am glad you pointed it out. I was just on this calculation last week and got really confused. Thanks.

chantelwetzel-noaa (approved):

Issue #516, opened 2021-04-13
automate processes like updating documentation, style, etc?


k-doering-NOAA:
I see there is a way to allow [github actions to push commits to a repository](https://github.com/actions/checkout#push-a-commit-using-the-built-in-token). I wonder if it would be worth it to try to automate some common workflows,?

For example, I see in the tidyr package, there is a [workflows setup](https://github.com/tidyverse/tidyr/blob/master/.github/workflows/pr-commands.yaml) that look for a certain issue comment on pull requests, and if found, will [use styler to style the pr automatically](https://github.com/tidyverse/tidyr/pull/1110#issuecomment-814769637) . This seems helpful if all new development is occurring through pull requests.

I think another option would just be to run devtools::document() and perhaps styler and the rm_dollar_sign function on a schedule (perhaps weekly?). I imagine we could do this directly on the main branch, or maybe have a branch made and submitted as a PR to review, if we want to review the differences just in case...

Thoughts if this would be useful/worth pursuing?

iantaylor-NOAA:
I think automatically running `devtools::document()` on pull requests seems very reasonable. On the other hand, I worry that `r4ss::rm_dollar_sign()` and `styler::style_pkg()`, while really valuable, could cause confusion if someone creates a PR, and then further changes are suggested before merging, but the code has changed from what got committed due to the changes those function make. So a weekly schedule, or an option to just run an an action on demand, might be the better approach. But I'm also a little confused by the "look for a certain issue comment" part. If that's something that only routine contributors to r4ss would add, fully understanding it's impact, then I'm good with all 3 steps being part of the process for every PR.

k-doering-NOAA:
Thanks @iantaylor-NOAA !

> On the other hand, I worry that r4ss::rm_dollar_sign() and styler::style_pkg(), while really valuable, could cause confusion if someone creates a PR, and then further changes are suggested before merging, but the code has changed from what got committed due to the changes those function make. So a weekly schedule, or an option to just run an an action on demand, might be the better approach.

Absolutely, it would be confusing!

>  But I'm also a little confused by the "look for a certain issue comment" part. If that's something that only routine contributors to r4ss would add, fully understanding it's impact, then I'm good with all 3 steps being part of the process for every PR.

I'm not sure if there are controls over who can trigger the action and I agree it doesn't really make sense to allow anyone to trigger it whenever they want on a PR.

[...]

PR #570, opened 2021-11-18
build, #516: add workflow to document and style

k-doering-NOAA:
Addresses #516. Automates running `devtools::document()` and `styler::style_pkg()`.

Right now, this is set up to run on every push to main, but we could modify this to run on a schedule (e.g., weekly), if that seems better to @iantaylor-NOAA.

Also, @iantaylor-NOAA , I know today is a focus day, please feel free to let this sit.


iantaylor-NOAA (approved):
Looks good to me. Let's see how it works--we can always refine in the future.
Thanks @k-doering-NOAA.

Issue #566, opened 2021-11-03
relative spawning biomass plots expected value and CI's are different


k-doering-NOAA:
I think maybe the CI's are not being calculated correctly? Using [this model](https://github.com/nmfs-stock-synthesis/user-examples/tree/main/model_files/simple_long) and ss3 v 3.30.18.

![screenshot of relative spawning biomass](https://user-images.githubusercontent.com/48930335/140079774-8ce2114a-5a7e-43ff-8659-d2f7d5f9fe54.png)

chantelwetzel-noaa:
Having worked on this section of the code recently, I suspect the actual issue is due to using option 2, Depletion basis:  2=rel SPBmsy, where the code that creates the plot is not grabbing the SPBmsy and dividing the relative spawning biomass by this value.  I would recommend looking at the changes done in the "depl_basis_plot" branch which deals with  other depletion scaling issues when creating this figure.  If we can ensure the existing changes in the "depl_basis_plot" branch are robust to various model versions, I think adding a fix for the SPBmsy should be relatively easy.

k-doering-NOAA:
Thanks @chantelwetzel-noaa! This model is using depletion option 2, I will look more into it!

iantaylor-NOAA:
I can help with this tomorrow but @k-doering-NOAA, I welcome you looking into it more.

I thought this was all resolved in https://github.com/r4ss/r4ss/issues/5, but clearly not.

`SSplotTimeseries()` may need a change similar to this recent change in `SSplotSPR()`, https://github.com/r4ss/r4ss/commit/c07f0bd0ad58775dc21df35c9dd561a5780c4574, where the `btarg` and `sprtarg` inputs specified in the forecast file aren't relevant when for some starter file options related to Depletion basis and SPR report basis.

k-doering-NOAA:
Thanks @chantelwetzel-noaa and @iantaylor-NOAA , I finally tried out the depl_basis_plot branch and the plots now look like this:
![image](https://user-images.githubusercontent.com/48930335/144633116-731beab0-76ed-45a8-9b6f-1b4198608e49.png)

I still don't understand why the first plot has different values than the second, though?

chantelwetzel-noaa:
Looking at these plots, my guess is that the first without uncertainty is still not grabbing the correct denominator, while the plot with uncertainty is.  I can take a look at the code in this branch to see where the issue may be.

k-doering-NOAA:
[...]

PR #580, opened 2021-12-07
Depl basis plot

k-doering-NOAA:
Thanks @chantelwetzel-noaa for providing fixes in this branch that corrects the denominator in the relative SSB plots.

I suggest this be merged into main, unless you are planning to make more commits to the branch in the short term (feel free to close this PR if you would rather this code not be merged into main yet).


chantelwetzel-noaa:
@k-doering-NOAA Thank you for following up on this.  I had not yet submitted a merge request since the fraction unfished plot when alternative denominators are requested (e.g, SBmsy, 0.50) for multi-area models has not been fixed yet.  The function works but creates non-sensical figures for fraction unfished without uncertainty.  The code in the main branch has a similar issue, so it does result in a degradation of performance but is still not 100%. If the team (@iantaylor-NOAA and @kellijohnson-NOAA ) are comfortable with merging this branch as-is then I will go ahead and do the review and merge.

k-doering-NOAA:
Thanks @chantelwetzel-noaa ! I only saw your discussion with Ian after submitting the PR.

My opinion is if something is better than main and fixing the whole problem may not happen for a while (not sure if this is the case or not!), merging the partial solution earlier is better then waiting (you could even add a warning about the problem, if it seems quick to do and the solution won't happen for a while).

  But, I agree, I'm happy to defer to what @kellijohnson-NOAA  and @iantaylor-NOAA  think.

Thanks again for your work on this! It was great to be able to get help with #566

iantaylor-NOAA:
I can take a little time today to look at what would be required for multi-area models and add to the documentation. If I'm not able to make quick progress I'll add a warning instead and go ahead and merge what we have.

iantaylor-NOAA:
Plots for multi-area models are fine when the depletion basis is relative to B0. In other cases, the axis label needs changing, but I think the plot is still OK. Cases of multi-area models with relative spawning output related to Bmsy are surely few enough that it's better to just merge and get this working for the larger set of single area models and I'll open a separate issue to track fixes to the rare cases that still need work.

Issue #514, opened 2021-04-08
rename  nondescriptive list item items created in SS_readforecast


k-doering-NOAA:
https://github.com/r4ss/r4ss/blob/main/R/SS_readforecast.R#L271-L273

```
forelist <- add_elem(forelist, "Forecast_loop_control_3")
forelist <- add_elem(forelist, "Forecast_loop_control_4")
```

These 2 items are named in a way that it is hard to tell what they are. Consider renaming to more descriptive names like `Forecast_Recruitment_Option`, `Forecast_Recruitment_Val` (but perhaps shortened?).  There is also "Forecast_loop_control_5", but this is a placeholder value for now.

Maybe these could be at first "soft deprecated" where both items are created by SS_readforecast(), but only the better named one is written out, with a warning if the 2 duplicate list items are not equal when reading out the object with SS_writeforecast()

iantaylor-NOAA:
Thanks @k-doering-NOAA.
Those variables were named [10 years ago](https://github.com/r4ss/r4ss/blame/f1ad95ab0a5fa9b1b17b34967fdd38639b10e102/R/SS_readforecast.R#L60-L62) when they were all placeholders, and it would indeed be good to update them.

I doubt that anyone is using them in a way that would be impacted by simply renaming the variables rather than soft deprecating. But if soft deprecating is easy, that plan sounds good too. I think the longer names you propose seem fine.

PR #584, opened 2021-12-13
#514: rename 2 elements in forecast file read/write

k-doering-NOAA:
Breaking change: rename Forecast_loop_control_3 to fcast_rec_option and FOrecast_loop_control_4 to fcast_rec_val. Users will be given a warning if the elements with the old names are not null.

Addresses #514 .


iantaylor-NOAA (approved):
Looks good to me, thanks @k-doering-NOAA

PR #587, opened 2021-12-17
fix: number of tag pars to read in ctl file

k-doering-NOAA:
Fixes an assumption in SS_readctl about how many lines of tagging params to read in 3.30. It should only be a parameter for each fleet of type 1 or type 2  ONLY( not including survey fleets).


k-doering-NOAA:
I merged this in, since it was a pretty simple fix and all the tests are passing

PR #588, opened 2021-12-17
change: address change in data new files

k-doering-NOAA:
In 3.30.19, SS3 will split data.ss_new instead into a file to echo back inputs (data_echo.ss_new), an expected values data file (if asked for), and a file for each bootstrap fle asked for. This PR allows SS_output and copy_SS_inputs to support data new files being named either data.ss_new or data_echo.ss_new.

### DIFF
--- a/R/copy_SS_inputs.R
+++ b/R/copy_SS_inputs.R
@@ -102,8 +102,12 @@ copy_SS_inputs <- function(dir.old = NULL,
       to = file.path(dir.new, starter[["ctlfile"]]),
       overwrite = overwrite
     )
+    # check data new file could be data.ss_new or data_echo.ss_new
+    dat_name <- 
+      ifelse(file.exists(file.path(dir.old, "data.ss_new")), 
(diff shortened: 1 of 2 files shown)


R/copy_SS_inputs.R
```diff
@@ -102,8 +102,12 @@ copy_SS_inputs <- function(dir.old = NULL,
       to = file.path(dir.new, starter[["ctlfile"]]),
       overwrite = overwrite
     )
+    # check data new file could be data.ss_new or data_echo.ss_new
+    dat_name <-
```
kellijohnson-NOAA, line 106: In the other function on line 1452you call it `datname` instead of `dat_name`. Also, this is really trivial but it would be good to start making helper functions, such as find_datafilename <- function(dir) {} rather than having the same code in two places.

k-doering-NOAA: thanks! I made both functions use datname, and also made the code into a helper function

PR #593, opened 2021-12-22
fix(get_dat_new_name): if was changing to NA

### LINKED ISSUE #592: SS_output crash at call of get_dat_new_name
When I called SS_output as
``
> out <- r4ss::SS_output(dir=".", covar = F)
``
SS_output stopped if data.ss_new exist but data_echo.ss_new  as

``
Finished dimensioning
Reading data.ss_new (or data_echo.ss_new) for info on Dirichlet-Multinomial parameters
Char version is  3.30
Numeric version is  3.3
Running SS_readdat_3.30
Error in file(con, "r") : cannot open the connection
In addition: Warning messages:
1: In get_dat_new_name(dir) :
  Neither data.ss_new or data_echo.ss_new found in dir .
2: In file(con, "r") : cannot open file './NA': No such file or directory
``

If I called get_dat_new_name directly, I got 

``
> r4ss:::get_dat_new_name(dir=".")
[1] NA
Warning message:
In r4ss:::get_dat_new_name(dir = ".") :
  Neither data.ss_new or data_echo.ss_new found in dir .
``

On the other hand 
``
> file.exists(file.path(".", "data.ss_new")
+ )
[1] TRUE
``
hence data.ss_new exists

According to the source codes of get_dat_new_name. 
"data.ss_new" is set to datname  if the file data.ss_new exists in dirt.
But then datname is reset if the file data_echo.ss_new does not exist.

One potential solution may be to let skip the latter part if datname is set to 
"data.ss_new" by "ifelse".  But I am not sure if this change is appropriate.   

``

kellijohnson-NOAA:
close #592 using dir(pattern) and ifelse

The pattern matching is done with `[e]*[c]*[h]*[o]*` which means 0+ instances of those characters in square brackets. Thanks to @k-doering-NOAA for mentioning what happens when the length of the match is nothing. The call to `tail()` is still fine because it returns `character(0)` and the `ifelse()` call changes it to `NA`. The `tail()` is needed in case both data.ss_new and data_echo.ss_new both exist. The returned value is alphabetized so the latter, which is the newer file, would always be the last value.

### DIFF
--- a/R/helper_fxns.R
+++ b/R/helper_fxns.R
@@ -15,7 +15,7 @@
 #'
 get_dat_new_name <- function(dir) {
   datname <- tail(
-    dir(path = dir, pattern = "data_*e*c*h*o*\\.ss_new"),
+    dir(path = dir, pattern = "data_?e?c?h?o?\\.ss_new"),
     1
   )
   ifelse(length(datname) == 0, NA, datname)
--- a/R/helper_fxns.R
+++ b/R/helper_fxns.R
@@ -4,17 +4,19 @@
 
 #' Get the name of the data .ss_new file in a directory
 #'
+#' In previous versions of Stock Synthesis,
+#' the file new data file was named `data.ss_new`.
+#' `_echo` was added to the name when the file was parsed into three separate
+#' files.
 #' @param dir Relative or absolute path to a directory
 #' @return A string with the name of the data .ss_new file. If not found, will
-#'  be NA.
+#'  be NA. Both of strings are searched for using `dir(pattern = )` and
+#'  if both exist, then `data_echo.ss_new` is returned.
+#'
 get_dat_new_name <- function(dir) {
-  datname <-
-    ifelse(file.exists(file.path(dir, "data.ss_new")),
-      "data.ss_new", "data_echo.ss_new"
-    )
-  if (!file.exists(file.path(dir, "data_echo.ss_new"))) {
-    datname <- as.character(NA)
-    warning("Neither data.ss_new or data_echo.ss_new found in dir ", dir)
-  }
-  datname
+  datname <- tail(
+    dir(path = dir, pattern = "data_*e*c*h*o*\\.ss_new"),
+    1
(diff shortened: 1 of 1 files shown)


kellijohnson-NOAA:
Looks like ubuntu does not like something 🤕 though it appears that it just cannot load {remotes}.

@k-doering-NOAA can you point me to the documentation for `?` instead of `*`? I am guessing this suggestion is great, I just did not know about it.

kellijohnson-NOAA:
[Linux documentation for grep](https://man7.org/linux/man-pages/man1/grep.1.html) explains the ?, which I do think is a great idea. Thanks @k-doering-NOAA.
```
   Repetition
       A regular expression may be followed by one of several repetition
       operators:
       ?      The preceding item is optional and matched at most once.
       *      The preceding item will be matched zero or more times.
       +      The preceding item will be matched one or more times.
       {n}    The preceding item is matched exactly n times.
       {n,}   The preceding item is matched n or more times.
       {,m}   The preceding item is matched at most m times.  This is a
              GNU extension.
       {n,m}  The preceding item is matched at least n times, but not
              more than m times.
```

k-doering-NOAA:
I always use this [regex cheatsheet](https://paulvanderlaken.files.wordpress.com/2017/08/r-regular-expression-cheetsheat.pdf) . The "quantifiers" section shows ? (matches at most 1 time).

FYI, I replicated the original bug by running SS_output on [this version of hake](https://github.com/nmfs-stock-synthesis/ss-test-models/tree/main/models/Hake_2018). I then tested on this branch, and it indeed fixes the issue! Thanks for working on it.

As for ubuntu, I saw this issue on the stock synthesis repo this morning, too, so I think it is something funny with not finding remotes, like you said - given the other runs are passing, I think it can be ignored.

k-doering-NOAA (approved):

k-doering-NOAA:
Tested with the new change and it still works great! Thanks @kellijohnson-NOAA .

PR #595, opened 2021-12-22
Get version information

kellijohnson-NOAA:
The start of a pull request to standardize getting the version of SS3.

Still need to look at the following files:
- [ ] SS_output.R
- [ ] SS_plots.R
- [ ] SSexecutivesummary.R
- [ ] SSplotCatch.R
- [ ] SSplotCohortCatch.R


kellijohnson-NOAA:
I am wondering if this is a mistake
https://github.com/r4ss/r4ss/blob/a532b8fb5aaa97625145a8bd5d848a90591be733/R/SS_plots.R#L466
because all other calls to `substr(SS_version)` in other files truncate the text at 8. Also, should we just add a full numeric version or minor version to the list returned by `SS_output`?

iantaylor-NOAA:
> I am wondering if this is a mistake
> 
> https://github.com/r4ss/r4ss/blob/a532b8fb5aaa97625145a8bd5d848a90591be733/R/SS_plots.R#L466
> 
> 
> because all other calls to `substr(SS_version)` in other files truncate the text at 8. Also, should we just add a full numeric version or minor version to the list returned by `SS_output`?

Yes, @kellijohnson-NOAA, definitely a mistake--exactly the sort of reason why it's good to standardize this process. 

Blame indicates that the difference has been around for at least 10 years, but that line seems to be only used when writing the plots to PDF which almost nobody does anymore and the impact would have been trivial anyway.

kellijohnson-NOAA:
@iantaylor-NOAA what do you think about removing the `version` argument from functions like `r4ss::SS_readdat()`? This would mean that when an old data file is being read in the user would have to open the file and add something like `#V2.00` to the first line of the file otherwise it would not be readable by `r4ss::SS_readdat()`. This would not be an issue for new files because they always have the version number. I am happy to continue developing the code either way you think is best.

iantaylor-NOAA:
Yes, let's get rid of the version argument. Users will also be responsible for updating the version number if they manually edit files created by a different version, but I think that's fine too.

Also, I forgot to comment on your second question from a week ago: "should we just add a full numeric version or minor version to the list returned by SS_output?"
I think anything would be an improvement over the current setup which is a total mess, with `SS_version`, `SS_versionshort`, `SS_versionNumeric`, all returned by SS_output(). `SS_versionNumeric` was intended to replace `SS_versionshort` (where long ago there were statements like `if SS_versionshort %in% c("3.21", "3.22", "3.23")` which had to be extended each time a new version came out. But there remain a few instances where that variable is used that can be updated.

I see that there are tools for dealing with semantic version numbers in R such as this package, https://github.com/zmarko/semver, but I'm not sure how well it work work for our system with 4 numbers.

kellijohnson-NOAA:
This pull request is still not complete but I wanted to get what I had updated before I go silent for a few weeks because of other priorities. Happy to have other contributors on this effort if someone wants to complete this sooner than I will have time to come back to it.

iantaylor-NOAA:
Thanks @kellijohnson-NOAA!
Just to clarify, are you saying we should leave this PR open until you are able to return to it in a few weeks and complete the work, or would you like others to try to pick up where you left off?

kellijohnson-NOAA:
Either is fine with me Ian. I pushed what I had so people could pick it up or make comments. I figured pushing what I had was better than it just sitting on my computer for a few weeks even if nothing happens.

k-doering-NOAA:
I converted this to a "draft" pr as  a reminder that it is not yet finished

PR #604, opened 2022-01-23
Fix a bug reading in ss.log

cgrandin:
Somewhere in new changes with SS, the ss.log file has the 'size' rows appearing on different line numbers. This removes hard-coded line numbers and does a regexp match on lines that begin with 'size'. There is also now an error message if these rows are not found.

### DIFF
--- a/R/SS_output.R
+++ b/R/SS_output.R
@@ -574,7 +574,11 @@ SS_output <-
       }
     }
     if (length(logfile) == 1 && file.info(file.path(dir, logfile))$size > 0) {
-      logfile <- read.table(file.path(dir, logfile))[, c(4, 6)]
+      logfile <- readLines(file.path(dir, logfile))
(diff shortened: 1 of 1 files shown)


cgrandin:
I can't remember which model it was, so I can't find it now but there was some text about gradient on the first line and then a couple of blank lines and then the correct lines like this:

size of file gradfil1.tmp = 0
size of file gradfil2.tmp = 0
size of file varssave.tmp = 0
size of file cmpdiff.tmp = 0

It's just that the hard coded row numbers 4 and 6 were off by one or two. It's much better to use grep in case this happens again

iantaylor-NOAA:
Thanks @cgrandin. 
I think the ss.log file is created by ADMB without modification, so perhaps this was related to different versions of ADMB or different operating systems.  Regardless, of the cause, the change is a clear improvement.

Issue #617, opened 2022-02-15
common utils.R 


MOshima-PIFSC:
In yesterday's discussion with the r4ss and ss3diags teams, Ian mentioned possibly creating a separate utils package for both packages to use. I like the idea but a separate, third package might be difficult to maintain. What about a common utils.R script where we keep basic functions? And then, if possible, we could set up a Github action so that if changes were pushed to the file from one repo it would be pushed to the other as well. I'm not sure exactly how to set that up but maybe [something like this](https://github.com/marketplace/actions/push-a-file-to-another-repository). 

Functions that are already in the ss3diags utils.R are:
*  sspar (used to make multiplots) 
* SSdiagsTime2Year (to convert non-annual into annual time steps)
* rc (same as r4ss::rich.colors.short)
* legendfun

Some candidate functions from r4ss might be:
* rich.colors.short (we can change rc function in ss3diags to rich.colors.short)
* matchfun
* matchfun2
* pngfun (ss3diags uses this function too but it is not in the utils yet)
* get_dat_new_name 

This was just an idea I had. Would be interested to hear others thoughts on this!
@efletcherPIFSC @fecor21 @Henning-Winker

k-doering-NOAA:
What do you think about just putting the functions in r4ss, @MOshima-PIFSC? I think this makes sense given that ss3diags already depends on r4ss. Changes to accomodate ss3diags could just be pushed directly to r4ss.

k-doering-NOAA:
Meant to say, I also like the idea of reducing the duplicated code!

MOshima-PIFSC:
@k-doering-NOAA I'm totally fine with that idea too!

k-doering-NOAA:
@MOshima-PIFSC great! Would you (and/or Eric) feel comfortable forking r4ss and submitting a PR for the changes you would like? That way @iantaylor-NOAA can take a look and make sure it doesn't break anything in r4ss.

Otherwise, I'm happy to coordinate with you and figure out how to best accomplish this.

roliveros-ramos:
My two cents:
Create utils-r4ss.R and utils-ss3diags.R, so one file is always a pasive copy in the other repo. While the automatic merge of git is quite good, non overlaping files can save some problems later.

iantaylor-NOAA:
Thanks for all the ideas. 
[...]

PR #621, opened 2022-02-17
r4ss common utility functions (helper_fxns.r)

efletcherPIFSC:
PR includes feature requests to include in `helper_fxns.R` from #617:

-  sspar (used to make multiplots)
-  SSdiagsTime2Year (to convert non-annual into annual time steps)
-  legendfun

-  rich.colors.short 
-  legendfun => add_legend (Adds legend to plots)
-  matchfun => match_report_line (Extract line number from Report.sso by keyword)
-  matchfun2 => match_report_table (Extract table from Report.sso by keyword)
-  pngfun => save_png (ss3diags uses this function too but it is not in the utils yet)

### DIFF
--- a/R/helper_fxns.R
+++ b/R/helper_fxns.R
@@ -162,7 +162,7 @@ SSdiagsTime2Year = function(ss3out,time.steps=0.25,end.time){
 #' 
   pngfun <- function(file, caption = NA){
     
-    if(filenameprefix){
+    if(exists("filenameprefix") == TRUE){
       # if extra text requested, add it before extention in file name (argument in ss3diags)
         file <- paste0(filenameprefix, file)
     }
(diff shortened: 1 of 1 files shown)


iantaylor-NOAA:
Thanks for the Pull Requestion @efletcherPIFSC.

Here are three minor suggestions/questions:
First, save_png() was added in my earlier Pull Request #620 so can be removed from this one. Calls to this function (previously `pngfun()` in ss3diags will need to be expanded to provide all the arguments like `pwidth` that were previously available locally within each function.

Second, renaming `matchfun()` and `marchfun2` as you've suggested makes lots of sense, but I've realized that those functions are currently only used within `r4ss::SS_output()`. Pulling them out of that function into helper_fxns.R will require providing the `obj` argument explicitly since `rawrep[,1]` used by `matchfun()` and `rawrep` used by `matchfun2()` don't exist outside of `SS_output()`. Would it make sense to just rename those functions within `SS_output()` or do you envision a use for those functions in other places? If there are things in Report.sso that ss3diags needs but doesn't have access yet, we should just expand `SS_output()` to get them.

Lastly, you might try running `devtools::document()` to generate the .Rd documentation files associated with the functions in the revised helper_fxns.R file (this also gets run automatically by a github workflow, but it can be useful to know if there are any issues with the files that it generates before merging the Pull Request.

PR #626, opened 2022-03-07
617 add legend

MOshima-PIFSC:
I added arguments: 
legendloc, legendorder, legendcol, legendcex, legendsp, col, pch, lty, and lwd.
Also @export was added to save_png


iantaylor-NOAA:
Looks good @MOshima-PIFSC. Thanks for helping tick off another item on the to-do list at #617.

PR #629, opened 2022-03-08
Minor fixes to save_png fun and add_legend

MOshima-PIFSC:
Added filenameprefix argument to save_png, and type = "l" to add_legend


iantaylor-NOAA:
Thanks for these changes @MOshima-PIFSC. The addition of `filenameprefix` to `save_png()` should allow this to work within SSplotComparisons() as well (where currently it has a separate version of that function).

MOshima-PIFSC:
Great!

PR #635, opened 2022-03-27
Replace write.fwf.r with write_fwf4.r

yukio-takeuchi:
Replaced write.fwf in SS_writectl/dat_3.24/3.30 with a new function write_fwf4.
write_fwf4 is mostly behave similar to write.fwf except for 
1. Default argument of na is "NA" instead of ""
2.include new arguments (checkError, checkNA and checkInfty)
   if checkNA is T, check if NA (NaN etc) is found and stop execution.
   if checkInfty is T, check if infinity is found and stop execution.
   checkError is a shortcut. It enable both. 
   Thse additional error check may allow to avoid situation 
   a. blank is printed out if a value is NA
   b. Inf is printed out if a value is infinity
3. behavior If a column vector contains positive and negative
    write.fwf prints extra digits for positive, while write_fwf4  respect digit specified.


iantaylor-NOAA (commented):
The changes all look good to me and I like the formatting without the extra significant digits much better. Thanks @yukio-takeuchi for putting this pull request together.

@k-doering-NOAA, since you've spent the most time with these read/write functions lately, could you take a quick look before we merge?

k-doering-NOAA:
Looks good! All the read/write tests are passing, so it looks like it is working as expected. Thanks @yukio-takeuchi !

Issue #563, opened 2021-10-25
add functions to read/write input files all at once


k-doering-NOAA:
A while back @iantaylor-NOAA mentioned there were some helpful functions created in lingcod that might be good to have accessible in r4ss. 
-  https://github.com/iantaylor-NOAA/Lingcod_2021/blob/main/R/get_inputs_ling.R reads in all the input files for a Stock Synthesis model. It would be helpful to do this generally in a directory.  Maybe call this `SS_read_inputs()`?
- There could be a similar function writing all at the same time, maybe `SS_write_inputs()`

kellijohnson-NOAA:
Do you think that we should shorten the names to `SS_read` and `SS_write` or would this make things difficult for autocomplete features of IDEs and text editors? I can start a branch with a version of the read function that I already started.

kellijohnson-NOAA:
@k-doering-NOAA I think this is complete, right?

k-doering-NOAA:
Yes, thanks @kellijohnson-NOAA !

PR #638, opened 2022-04-01
add functions to read/write input files all at once

iantaylor-NOAA:
This PR should complete the addition of `SS_read()` and `SS_write()`  functions proposed in #563 which wrap up the 5 individual read/write functions for the numerous applications where its nice to deal with all files together.

@kellijohnson-NOAA, I made some minor changes to the `SS_read()` function you created back in November. Feel free to modify further as you wish. And the new `SS_write()` can also be changed anyway that you wish.

FYI, I rebased the `SSfilewrap` branch onto main in order to run the current github actions before creating this PR.


R/SS_read.R
```diff
@@ -28,11 +27,15 @@
...
 #' @export
 #' @seealso
-#' * [SS_readstarter()], [SS_readdat()], [SS_readctl()], and [SS_readforecast()]
-#'   are used to read in the input files.
+#' * [SS_write()] can be used to write the input files using the list
+#'   created by this function.
+#' * [SS_readstarter()], [SS_readdat()], [SS_readctl()],
+#'   [SS_readforecast()], and [SS_readwtatage()] are used by this
+#'   function to read in the input files.
 #' * [SS_output()] to read in equivalent SS output files.
```
kellijohnson-NOAA, line 39: "SS output files" should be "SS3 output files" here.

R/SS_write.R
```diff
@@ -0,0 +1,110 @@
...
+      dir = dir,
+      overwrite = overwrite,
+      verbose = verbose
+    )
+  }
+  # message noting that all files have been written
+  if (verbose) {
+    message("Wrote all input files to ", dir)
+  }
+}
```
kellijohnson-NOAA, line 110: wants a new line at the end of the file.

R/SS_write.R
```diff
@@ -0,0 +1,110 @@
...
+#' inputs$ctl$MainRdevYrFirst <- 1980
+#'
+#' # write the files to a new folder within the source directory
+#' SS_write(inputs = inputs,
+#'          dir = file.path(inputs$dir, "modified_inputs"))
+#' }
+
+SS_write <- function(inputs,
+                     dir = NULL,
+                     files = c("dat", "ctl", "start", "fore", "wtatage"),
```
kellijohnson-NOAA, line 50: Seems like you could get rid of this argument and just write the files that are in the list. This would force users to always write the starter file though b/c you could never remove that element from the list. Anyway, I think that the input argument `files` makes things overly complicated and if users want to write just a few rather than all they should just use the individual functions.

R/SS_write.R
```diff
@@ -0,0 +1,110 @@
...
+  # create directory if not already there
+  if (!is.null(dir)) {
+    if (!dir.exists(dir)) {
+      if (verbose) {
+        message("Creating new directory: ", dir)
+      }
+      dir.create(dir)
+    }
+  } else {
+    dir <- ""
```
kellijohnson-NOAA, line 62: Why make this a hidden input? Couldn't you just make the default ""? It seems like it would be more transparent to the user that way.

R/SS_write.R
```diff
@@ -0,0 +1,110 @@
...
+#' # modify the control file (turn off early recdevs and change range of yrs)
+#' inputs$ctl$recdev_early_phase <- -abs(inputs$ctl$recdev_early_phase)
+#' inputs$ctl$MainRdevYrFirst <- 1980
+#'
+#' # write the files to a new folder within the source directory
+#' SS_write(inputs = inputs,
+#'          dir = file.path(inputs$dir, "modified_inputs"))
+#' }
+
+SS_write <- function(inputs,
```
kellijohnson-NOAA, line 48: nit: `inputs` made me think that you had to supply multiple things to this argument, where really you just need to supply a single list. I would have named it `inputlist` to be more consistent with `datlist`. I would be fine with leaving it, just my two cents.

kellijohnson-NOAA:
Thanks @iantaylor-NOAA for finishing this. I added a few comments based on things that popped into my head while looking at the code changes. Feel free to ignore all of them.

R/SS_write.R
iantaylor-NOAA: The workflow that I used for the `write_inputs_ling()` function which preceded this one was to read all inputs files at the start, then modify a subset and only write out the ones that got modified (e.g. for a particular sensitivity could be either control or data: https://github.com/iantaylor-NOAA/Lingcod_2021/blob/main/R/run_sensitivities.R). 

The arguments in the SS_write*() functions depend on the file being written ("outfile" is used for data and control files, while "dir" is used for starter and forecast since the names are fixed). So for me it requires less thought to just run this wrapper function each time.

I did simplify from `write_inputs_ling()` by removing control of the filenames under the assumption that the user would just modify the values in `inputlist$start`.

iantaylor-NOAA:
New functions don't impact the rest of the package, so no reason not to merge and make them available.
All but one of the suggested improvements from @kellijohnson-NOAA have been incorporated in e54c577.

R/SS_write.R
iantaylor-NOAA: Issue resolved via discussion and revisions to `SS_write()` in 2862de3. `SS_write()` will now write whatever elements are present in the list. It's not more efficient (may overwrite existing file with no changes), but it's simpler and easier to understand.

Issue #641, opened 2022-04-05
default of SS_writectl @version is not the same as the default of SS_writedat @version


kellijohnson-NOAA:
https://github.com/r4ss/r4ss/blob/17a16966e786820081685b05164f70656457b65e/R/SS_writectl.R#L23 defaults to NULL which then leads to r4ss attempting to determine the version. Where, `SS_writedat(version = "3.30")`. All other functions with the version argument have `version = NULL`. I am really not sure on which option is the best but I think having one default would be good.

iantaylor-NOAA:
I'm think that in the past it made sense to have these functions be smart where possible and figure out which version to use, but at this point, the use of version 3.24 should be so rare that anyone still doing can be asked to be really explicit about it.

I see `version = "3.30"` inputs for: `SS_output()`, `SS_parlines()`, `SS_profile()`, `SS_readforecast()`, and `SS_varadjust()`, whereas I only see `version = NULL` in `SS_writectl()`, `SS_readctl()`, and `SS_readdat()`.

It seems reasonable to change NULL to "3.30" in those three cases as long as we also add a warning when there's evidence that this might be wrong (e.g. modify the lines below to warn if `ctllist[["ReadVersion"]]` doesn't match the input:
https://github.com/r4ss/r4ss/blob/17a16966e786820081685b05164f70656457b65e/R/SS_writectl.R#L36-L39

kellijohnson-NOAA:
Great, I didn't want to specify a preference in my original issue, but having `version = 3.30` as the default will simplify code in {ss3sim}.

kellijohnson-NOAA:
@iantaylor-NOAA would you like me to submit a pull request for this?

iantaylor-NOAA:
Yes, that would be awesome. Thank you.

kellijohnson-NOAA:
Should the use of `version` be removed when calling `SS_readdat(version = "3.30")` within `SS_output()`?

https://github.com/r4ss/r4ss/blob/17a16966e786820081685b05164f70656457b65e/R/SS_output.R#L1457-L1460

iantaylor-NOAA:
@kellijohnson-NOAA, good questions.
* Regarding `SS_output()`: yes, remove the `version` input now that the default is provided by `SS_readdat()` instead. That line will only be called for 3.30 models anyway because DM parameters didn't exist in 3.24.
[...]

PR #643, opened 2022-04-06
Fix argversion

kellijohnson-NOAA:
Standardize the input argument `version = "3.30"` rather than having some use the `NULL` option.

I am not sure if the third commit that removes checks for `NULL` should be left in. I purposefully made it the last commit and separated it from everything else such that we can easily remove the commit from this pull request.

I did **NOT** find `version = NULL` in any of the tests, nor did I write any tests 👎 😢  One additional thing to think about is your request in #641 to update the check to see if the version in the list matches the argument. I have not addressed this as of yet, which is why this is just a draft. I think I will just make a stop() call if the values do not match. Let me know if you do not agree.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: r4ss
 Type: Package
 Title: R Code for Stock Synthesis
-Version: 1.43.1
+Version: 1.43.2
 Author: Ian G. Taylor, Ian J. Stewart, Allan C. Hicks, Tommy M. Garrison,
     Andre E. Punt, John R. Wallace, Chantel R. Wetzel, James T. Thorson,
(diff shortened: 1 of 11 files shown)


iantaylor-NOAA (approved):
Looks good. I see no reason to make it easy to profile 3.24 models when we can instead simplify code that's frequently being used.

kellijohnson-NOAA:
kk I am done. I changed to using `deprecate_stop()` instead of `deprecate_warn()` because it is unclear what to use as the default in this instance.

iantaylor-NOAA (approved):

kellijohnson-NOAA:
This pull request closes #641.

Issue #645, opened 2022-04-06
customize code cov to download models and exes first


k-doering-NOAA:
> The code coverage calculation doesn't seem to account for the pieces that require SS3 input files in inst/extdata/models: https://app.codecov.io/gh/r4ss/r4ss/compare/638/diff. Do you know if we can fix that?

Yes, you are absolutely correct. I bet we could include a step to download the models and then run codecov. BTW, looks like there are a lot of options I didn't know about for the covr package (https://covr.r-lib.org/)

_Originally posted by @k-doering-NOAA in https://github.com/r4ss/r4ss/issues/574#issuecomment-1087962225_

k-doering-NOAA:
Note that code coverage can be calculated locally for a package by using `covr::package_coverage()`. I found that downloading models and exes first increases codecov by ~4%, which is fairly significant.

Issue #574, opened 2021-11-22
Create function to download "models" directory and provide documentation


kellijohnson-NOAA:
https://github.com/r4ss/r4ss/blob/44cf139540e8d7e13fe89c69a05f8f7a55d2959e/tests/testthat/test-readwrite.R#L4
as well as 
https://github.com/r4ss/r4ss/blob/44cf139540e8d7e13fe89c69a05f8f7a55d2959e/tests/testthat/test-readwrite.R#L8
reference a directory called "models" in r4ss. This directory is not actually present in the r4ss repository but is something that developers download from [Stock Synthesis ss-test-models repository](https://github.com/nmfs-stock-synthesis/ss-test-models/tree/main/models). Being able to easily download those models, e.g., if you are setting up a new machine, to test the `SS_read*()` and `SS_write*()` functions and documentation in the test how to get the models would be good.
I think we should create a function in r4ss that downloads the models for you, that way in the test we can just call the function to download the models. This function could have a check to see if you already have the models directory in r4ss\inst\models it does nothing. I am currently looking into how to download a single directory. @k-doering-NOAA found how to download individual files. Perhaps {usethis} has something.

iantaylor-NOAA:
@kellijohnson-NOAA and @k-doering-NOAA, this seems like a great idea. Let me know how I can help.

iantaylor-NOAA:
I modified test-readwrite.R without remembering that the github actions don't run it because it requires inst/extdata/models. Running the test locally helped me sort out a bunch of issues (in #638) but it made me appreciate the day when we can test all those models via github actions.

k-doering-NOAA:
@iantaylor-NOAA, I think the r cmd check github actions should run it, because it includes a step to download the models: https://github.com/r4ss/r4ss/blob/main/.github/workflows/call-r-cmd-check.yml#L20, which links to this workflow: https://github.com/nmfs-stock-synthesis/workflows/blob/main/.github/workflows/r4ss-extra-tests.yml#L61-L71

iantaylor-NOAA:
@k-doering-NOAA this is helpful, thank you!

I still have a variety of questions about workflows that a meeting might address better than more github issue, but for future reference, here's a few:
- The code coverage calculation doesn't seem to  account for the pieces that require SS3 input files in `inst/extdata/models`: https://app.codecov.io/gh/r4ss/r4ss/compare/638/diff. Do you know if we can fix that?
[...]

PR #648, opened 2022-04-07
gha: calculate more accurate code coverage

k-doering-NOAA:
Addresses #645

As discussed in https://github.com/r4ss/r4ss/issues/574#issuecomment-1087962225 , code coverage is currently not accurate, because there are tests in r4ss that rely on executables and models. This adds a workflow where the ss3 exes and model files are downloaded before calculating code coverage to improve the accuracy.

Note this also switches to using ubuntu instead of osx to run the job; I think this should work fine, but if it doesn't we can switch the github runner to osx.


iantaylor-NOAA:
This is great. Thank you @k-doering-NOAA!
Significant increases for lots of important functions will help us better understand remaining areas where testing could be improved the most.

PR #659, opened 2022-04-26
Fix two morph

k-doering-NOAA:
Trying to plot results from this [two morphs model](https://github.com/nmfs-stock-synthesis/test-models/tree/main/models/two_morph) run with ss3.30.19 illuminated some issues (a missing space in the Report file, and issues creating plots when certain reported quantities aren't available. This PR provides fixes to allow `SS_plots()` to run.

### DIFF
--- a/R/SS_output.R
+++ b/R/SS_output.R
@@ -1886,8 +1886,7 @@ SS_output <-
           # temp fix  for 3.30.19
           if (length(grep("RECRUITMENT_DIST_TIMESERIES", recruit_dist_endyr[["Settle#"]])) == 1) {
             tmp_brk_line <- grep("RECRUITMENT_DIST_TIMESERIES", recruit_dist_endyr[["Settle#"]]) - 1
-            recruit_dist_endyr <- recruit_dist_endyr[seq_len(tmp_brk_line),]
-                                 
+            recruit_dist_endyr <- recruit_dist_endyr[seq_len(tmp_brk_line), ]
           }
         }
         # bundle original and extra tables into a list
--- a/R/SS_output.R
+++ b/R/SS_output.R
@@ -1883,6 +1883,12 @@ SS_output <-
           recruit_dist_endyr <- match_report_table("RECRUITMENT_DIST_endyr", 1,
             header = TRUE, type.convert = TRUE
           )
+          # temp fix  for 3.30.19
+          if (length(grep("RECRUITMENT_DIST_TIMESERIES", recruit_dist_endyr[["Settle#"]])) == 1) {
+            tmp_brk_line <- grep("RECRUITMENT_DIST_TIMESERIES", recruit_dist_endyr[["Settle#"]]) - 1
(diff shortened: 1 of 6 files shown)


k-doering-NOAA:
I think the failing mac job is not related to this PR; I think it might be due to a missing package, which may fix itself if we wait a day and rerun the job.

R/SSplotBiology.R
```diff
@@ -1581,7 +1581,12 @@ SSplotBiology <-
     } # end disable of time-varying growth for multi-season models
 
     # plot time-series of any time-varying quantities
-    if (24 %in% subplots) {
+    if(is.null(MGparmAdj)) {
```
kellijohnson-NOAA, line 1584: should this just be an if{}else{} that is part of the original if statement  below? That way users only get the message if 24 %in% subplots?

R/SSplotMovementRates.R
```diff
@@ -116,7 +116,12 @@ SSplotMovementRates <-
     } # end subplot 1
 
     # subplot 2: time-varying movement
-    if (2 %in% subplots) {
+    if(is.null(MGparmAdj)) {
```
kellijohnson-NOAA, line 119: Same comment as above about if{}else{}

kellijohnson-NOAA: Also, for both of these lack of figure messages, should we make a note in the documentation for the functions that these figures are not available?

kellijohnson-NOAA (approved):
Thanks k-doering-noaa for this pull request. Everything is very minor that I noted, except I think that users should be notified in the documentation of Figures that do not work when multiple morphs are present. In general we provide really good documentation for input arguments but not so much for functionality or what something returns. So, it might be good to notify users of this lack of functionality prior to running the function.

R/SS_output.R
```diff
@@ -1883,6 +1883,12 @@ SS_output <-
           recruit_dist_endyr <- match_report_table("RECRUITMENT_DIST_endyr", 1,
             header = TRUE, type.convert = TRUE
           )
+          # temp fix  for 3.30.19
```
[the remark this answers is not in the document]
k-doering-NOAA: I did already modify it in Stock Synthesis (in commit https://github.com/nmfs-stock-synthesis/stock-synthesis/commit/d618c3c0c6d58da9f532ed531149e1186b31b4e3), but this will still be needed for .19 models. I shouldn't have labeled this a "temp fix", I think it is a fine permanent fix for the missing space in .19 Report files.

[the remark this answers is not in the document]
kellijohnson-NOAA: That makes complete sense now. Thanks for the explanation. Maybe we should note the SS3 commit to fix the missing space in the comment instead of temp fix.

R/SSplotBiology.R
k-doering-NOAA: Yes, good call! I changed to using an if else statement for MGparmAdj

Issue #316, opened 2019-07-01
expand testing via testthat 


iantaylor-NOAA:
@kellijohnson-NOAA added new tests in https://github.com/r4ss/r4ss/blob/development/tests/testthat/test_profile.R of functions that are used by `SS_profile`. The `SS_profile` function itself was included in a set of tests in https://github.com/r4ss/r4ss/blob/master/tests/testthat/test_runs.R that require the SS executable and therefore were only run on local machines, not automatically at https://travis-ci.org/r4ss/r4ss. However, that `test_runs.R` script had issues with dependency on the `SSutils` package which I didn't take the time to sort out so commented out the whole thing. 

In general, expanding all the tests should make the package easier to maintain and cause fewer problems for users.

iantaylor-NOAA:
The `mergeSSutils` branch merges the 3 functions from [`SSutils`](https://github.com/r4ss/ssutils) into r4ss: `copy_SS_inputs.R`, `populate_multiple_folders.R`, and `run_SS_models.R`.

This solves the dependency on `SSutils` which caused a problem for the `test_runs.R` file noted at the start of this issue a year ago. This will now allow additional tests of functions such as `SS_doRetro()` and `SS_profile()` that require the SS executable.

The function `skip_if()` is used in `test_runs.R` to check for the presence of an executable in the example model directories within \r4ss\inst\extdata. On my computer the `testthat` subset of `devtools::check()` takes about 11 minutes to run all the tests while the Travis-CI version skips those that require an executable (because the .exe are excluded from the repository via `.gitignore`) and takes only about 2 minutes.

The `test_runs.R` file (or new similar files) could now be expanded to include a few additional tests at which point I think we could close this issue and open specific ones associated with particular tests as needed in the future.

Thanks @kellijohnson-NOAA and @k-doering-NOAA for help on this.

iantaylor-NOAA:
Use of `if (skipexe) { } else { }` to tests function with and without executable as @kellijohnson-NOAA has done in  https://github.com/r4ss/r4ss/commit/6c8d9b1edbaccbd44b35a3da6c2a88e58215589c is a good idea. We should copy that for other tests in the future.

k-doering-NOAA:
[...]

PR #663, opened 2022-04-28
Improve codecov by adding tests for functions that don't have any

k-doering-NOAA:
Related to #316 (note this issue was closed a while ago because it is open-ended and not helpful).

While looking at codecov.io for r4ss, I realized there were a number of files that don't have any coverage at all. I attempted to add tests for some of these functions that I know get use.

This PR should increase code coverage, for the functions `SS_mohnsrho`, `getADMBHessian`, `TSCplot`, `plotCI`, and `populate_multiple_folders`. 

I also made some changes to functions and documentation as noticed while working on adding these tests. In particular, there was an issue with the Mohn's rho calculation where the necessary output was not available. I added a check for this and skipped the particular calculations, but there may be a smarter way of dealing with this problem.

### DIFF
--- a/.Rbuildignore
+++ b/.Rbuildignore
@@ -11,3 +11,4 @@ docs
 ^jenkins
 ^\.github/
 ^\.github$
+^\.covrignore$
--- a/.covrignore
+++ b/.covrignore
@@ -1,0 +1,1 @@
+R/r4ss_logo.R
--- a/R/SS_doRetro.R
+++ b/R/SS_doRetro.R
@@ -2,7 +2,9 @@
 #'
 #' Do retrospective analyses by creating new directories, copying model files,
 #' and iteratively changing the starter file to set the number of years of data
-#' to exclude.
+#' to exclude. Note that there was a  bug for retrospectives in 3.30.01;
+#' the user should update their model to a newer version of Stock Synthesis to
+#' run retrospectives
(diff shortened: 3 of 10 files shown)


iantaylor-NOAA:
This is excellent, thank you @k-doering-NOAA!
I think that the next step to improve code coverage would be to include a larger set of test models. For instance, the `plotCI()` function is used for plots of mean body weight and discards, but the simple test models don't include either of those data types. Likewise, `SSplotBiology()` would have better coverage if a wtatage model were included in the test.

Thanks to you, r4ss is already being tested on a large set of models in the SS3 repo: https://github.com/nmfs-stock-synthesis/stock-synthesis/actions/workflows/call-test-r4ss-with-ss3.yml. Do you think there's a way to compute the code coverage from that test, or repeat the test here within the r4ss repo once a week or once a month or something like that?

k-doering-NOAA:
Yep, that is a good point about the larger set of test models! I'm working on a refactor of the test-models repository to try to cover more features with fewer models. So, I'll stick with that as a way to improve both SS3 and r4ss testing simultaneously :) 

We are getting "credit" for testing r4ss with our test-models, as the models are downloaded before calculating code coverage. That led to what I now understand is a pretty significant jump in coverage, of about 5% !

PR #675, opened 2022-06-09
fix(SS_writedat_3.30): Adds check for no mean size

kellijohnson-NOAA:
Users might erroneously have `d[["use_MeanSize_at_Age_obs"]] == 1`
when there are no observations. This happened to me in a
simulation when prior to sampling I had mean size, then
after sampling the data frame was NULL.
Thus, the new check ensures that the observations are not NULL
and if they are, then it sets d[["use_MeanSize_at_Age_obs"]] to 0.

@k-doering-NOAA did I use the `&&` appropriately here or should it be `&`?

### DIFF
--- a/R/SS_writedat_3.30.R
+++ b/R/SS_writedat_3.30.R
@@ -392,6 +392,9 @@ SS_writedat_3.30 <- function(datlist,
   }
 
   writeComment("#\n#_MeanSize_at_Age_obs")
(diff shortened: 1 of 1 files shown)


k-doering-NOAA (approved):
Looks good to me! 

Just for clarity, I think both && and & work in this case; the difference is && will always evaluate both statements even if the first one is FALSE, whereas & will not bother to evaluate the second statement if the first is FALSE.

PR #679, opened 2022-06-13
677 `SS_read()` allows a url

kellijohnson-NOAA:
Had to rework the check for non-existent files or files without any information in `SS_readstarter()` and `SS_readwtatage()` to accommodate a URL yet would still return `NULL` in the latter function when appropriate.

* Allows for a URL to `SS_read(dir=)`, which facilitates using `users-examples` in the Stock Synthesis organization.
* Increases documentation of `SS_read` functions by adding an example with weight-at-age data and providing some detailed information about `SS_readstarter()` that might be helpful to new users.
* Uses @template verbose instead of bespoke text.
* Adds `mustWork = FALSE` to normalize path because a URL will lead to a warning.
* Adds test using a URL
* Fixes formatting of line length and a few other minor formatting issues.

### DIFF
--- a/R/SS_read.R
+++ b/R/SS_read.R
@@ -10,7 +10,7 @@
 #' the control and data files.
 #'
 #' @param dir A file path to the directory of interest or a raw github URL (see
-#' example). The default value is \code{dir = NULL}.
+#' example). The default is the current working directory, `dir = getwd()`.
 #' @param ss_new A logical that controls if the `.ss_new` files or
 #'   the original input files are read in.
 #'   The default is to read the original files.
@@ -45,7 +45,7 @@
 #' # Read in an example from GitHub stored in user-examples,
 #' # wrapped in `dontrun` because it requires an Internet connection
 #' \dontrun{
-#' webexample <- SS_read(file.path(
+#' webexample <- SS_read(dir = file.path(
 #'   "https://raw.githubusercontent.com",
 #'   "nmfs-stock-synthesis",
 #'   "user-examples",
@@ -54,12 +54,9 @@
 #'   "simple_long_wtatage"
 #' ))
 #' }
-SS_read <- function(dir = NULL,
+SS_read <- function(dir = getwd(),
                     ss_new = FALSE,
                     verbose = FALSE) {
-  if (is.null(dir)) {
-    dir <- getwd()
-  }
 
   # Read in starter first to find the names of the input files
   start <- SS_readstarter(
--- a/R/SS_read.R
+++ b/R/SS_read.R
@@ -90,7 +90,7 @@ SS_read <- function(dir = NULL,
   # return a list of the lists for each file
   invisible(list(
     dir = dir,
-    path = normalizePath(dir),
+    path = normalizePath(dir, mustWork = FALSE),
     dat = dat,
     ctl = ctl,
     start = start,
--- a/R/SS_read.R
+++ b/R/SS_read.R
@@ -5,9 +5,9 @@
 #' Functionality comes from the `r4ss::SS_read*()` functions.
 #' This function simplifies the number of lines of code you need to write by
 #' using all of the read functions to read in the
-#' starter, control, data, and forecast files and maybe the weight-at-age file.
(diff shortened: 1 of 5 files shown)


k-doering-NOAA:
Thanks for working on this, @kellijohnson-NOAA! I really like the idea of expanding the inputs that `dir` can take. I tried adding a test illustrating how I think SS_read should work when dir is a URL and couldn't get it to pass locally...I pushed it up the branch so you could take a look. Am I doing something wrong?

tests/testthat/test-readwrite.R
```diff
@@ -83,3 +83,10 @@ test_that("empty files lead to NULL or error", {
     SS_readstarter("nonexistentfile.ss", verbose = FALSE)
   ))
 })
+
+test_that("ss_read works with a URL", {
+  skip_if_offline(host = "github.com")
+  list_objs <- SS_read(dir = 
+  "https://github.com/nmfs-stock-synthesis/test-models/tree/main/models/Simple")
```
kellijohnson-NOAA, line 90: You have to use raw user content. See the example in `SS_read()`. We should probably add something to the documentation about how to call from GitHub? 🤷

k-doering-NOAA: Ah, of course. I added a little bit to the documentation to hopefully guide the user to use the raw content!

R/SS_read.R
```diff
@@ -5,29 +5,28 @@
...
 #' which gives the full path.
-#' The remaining four elements are list objects from reading in
+#' The remaining five elements are list objects resulting from reading in
 #' the following input files:
 #' * data
 #' * control
 #' * starter
 #' * forecast
-#' * wtatage (will be NULL if not required by the model)
+#' * wtatage (will be `NULL` if not required by the model)
```
k-doering-NOAA, line 29: If a list object is `NULL`, it is often just dropped from the object. If there should defintely be a list object called wtatage even if wtatage is read in, it should be assigned `NA` instead.

R/SS_read.R
```diff
@@ -43,6 +42,18 @@
 #' inputs <- SS_read(
 #'   dir = system.file("extdata", "simple_3.30.13", package = "r4ss")
 #' )
+#' # Read in an example from GitHub stored in user-examples,
+#' # wrapped in `dontrun` because it requires an Internet connection
```
k-doering-NOAA, line 46: Is this a best practice if something requires an internet connection?

kellijohnson-NOAA: imo yes; but, I am not sure if this is a documented best practice, just what I do because sometimes I am testing packages without an internet connection.

R/SS_read.R
```diff
@@ -5,29 +5,28 @@
...
-#' starter, control, data, and forecast files.
-#' The starter file is helpful because it provides names for the
-#' control and data files.
+#' starter, control, data, and forecast files and if requested, the
+#' weight-at-age file. The starter file is helpful because it provides names for
+#' the control and data files.
 #'
-#' @template dir
+#' @param dir A file path to the directory of interest or a raw github URL (see
+#' example). The default value is \code{dir = NULL}.
```
kellijohnson-NOAA, line 13: @k-doering-NOAA I am wondering if we should take this opportunity to uncover the hidden argument of `dir = NULL` where it actually leads to `dir = getwd()`. Best practices suggests avoiding hidden arguments and just setting them in the function call.

R/SS_read.R
kellijohnson-NOAA: I think it is fine as of right now to leave this as NULL but maybe I can take a stab at changing the documentation to reflect what you said b/c knowing that behavior is helpful, or at least it was for me.

kellijohnson-NOAA: I re-read the documentation and I think it is fine as is.

R/SS_read.R
k-doering-NOAA: that is fine by me! I think that makes sense.

R/SS_read.R
k-doering-NOAA: Ok!

Issue #678, opened 2022-06-13
`faster` argument is still in tests for simple


kellijohnson-NOAA:
Here is the first of 12 instances of `faster` in this test file
https://github.com/r4ss/r4ss/blob/d44acd3820dbef8fa1a9f09fb7d30c7781a6f748/tests/testthat/test-simple.R#L185

where 5fd3197a0fd26555b972d5c5baa82b20ff1c1494 deprecated the `faster` argument. Thus, I think it should be removed from the tests. But, I am not sure if a few of the tests can also be removed because it appears as though it might be exclusively there to just test the `faster` argument itself.

PR #680, opened 2022-06-13
Remove faster from tests; deprecate echoall argument in SS_readdat* functions

k-doering-NOAA:
Addresses #678. Also extends work done in #671, which included deprecating unneeded arguments in the read/write functions. For some reason, the echoall argument was not removed from the SS_readdat* functions, so it is done in this pull request

### DIFF
--- a/R/SS_readdat.R
+++ b/R/SS_readdat.R
@@ -15,8 +15,7 @@
 #'  line of the file.
 #' @param verbose Should there be verbose output while running the file?
 #' Default=TRUE.
-#' @param echoall Debugging tool (not fully implemented) of echoing blocks of
-#' data as it is being read.
+#' @param echoall Deprecated.
 #' @param section Which data set to read. Only applies for a data.ss_new file
 #' created by Stock Synthesis. Allows the choice of either expected values
 #' (section=2) or bootstrap data (section=3+). Leaving default of section=NULL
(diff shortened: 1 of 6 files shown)


kellijohnson-NOAA (approved):
I am just wondering if we there are some calls that were printed to the screen with echoall that should maybe not be moved over to verbose? It just seems like an overwhelming amount of information is now printed to the screen where it was already a lot before.

kellijohnson-NOAA:
Thanks for the background @iantaylor-NOAA. I agree with removing all the old output from echoall and leaving verbose to show the progress of the function and allow users to preview summaries that they might not know about. That is, I do not think there is a lot of utility in printing read in data like the catches to the screen.

k-doering-NOAA:
@kellijohnson-NOAA , does this look better? For SS_readdat_3.30, I also took the time to refine the verbose msgs a bit to reflect their intended purpose (helping users track the function running, rather than printing output to the screen)

Also, IDK if we want to bump the version (just the patch?) relative to the main branch before merging in? Feel free to push a commit to do this, or I can - I should also bump the version on the lifecycle deprecation msg, too, fi that is the case.

kellijohnson-NOAA:
This looks great @k-doering-NOAA feel free to merge in whenever you are ready.

Issue #311, opened 2019-06-20
Using Roxygen template tags to document parameters used in multiple functions


k-doering-NOAA:
In r4ss, the same parameters are often used across several functions (for example, see the common parameters in ```SS_readctl()```, ```SS_readctl_3.30```, and ```SS_readctl_3.24```). It may be helpful to use template tags ([see this stack overflow thread](https://stackoverflow.com/questions/15100129/using-roxygen2-template-tags)) to define parameters instead of adding ```@param some_param description``` for each function that uses the parameter. For example, one template file (say ```ctltemp.R```) could be created that contains all the common parameters across the SS_readctl functions. Then, in the R oxygen header lines where the functions themselves are defined, ```@template ctltemp``` can be called instead of writing out each parameter name and its definition.

This is probably low priority and doesn't improve the functionality of r4ss, but it may make managing documentation simpler in the future. Also, if there is a better way to do this, or if there are good reasons to leave the Roxygen header lines as-is, please comment on this issue. 

One reason I could think of to leave it the way it is is that some developers may not know about template tags or where to find the templates, and it may appear to them that there are parameters missing documentation in the Roxygen header lines, or at the very least they can't easily glance at the uncompiled documentation and determine if it should be updated or not.

iantaylor-NOAA:
Thanks @k-doering-NOAA. I think this is a great idea. There are surely numerous inconsistencies in the definitions of the input parameters, so bringing some standardization is a good idea. 

It's probably also a good idea to clean up inconsistent inputs. For instance, SS_plots takes a `printfolder` input which needs to be a relative path whereas `SSplotComparisons` takes a `plotdir` which could be an absolute path or relative to the current working directory. Maybe it's good to keep them separate, since they aren't likely to share a definition anyway. In other cases, the inconsistency is between the `SS_plots` inputs and the graphics parameters associated with `par()`, such as `ptsize` instead of the standard `pointsize`.

I don't know if there's an easy way to make a list of parameters that are used in multiple functions, but do you think would be the right way to start?

k-doering-NOAA:
@iantaylor-NOAA making a list to start does sound like a good way to go, so that it could be shared (perhaps on this thread or in the Wiki? ) before changes to the documentation are made. I'm not aware of an easy way to do this other than manually, but perhaps someone else has an idea?

[...]

Issue #687, opened 2022-06-15
subplots and subplot both used as variables in plotting functions


k-doering-NOAA:
Wile working on #311, I noticed that both `subplots` and `subplot` are used as arguments in the SSplot* functions, though as far as I can tell, they mean the same thing (at least in most cases). To reduce confusion, I think it would be great to change all of these to subplots and soft deprecate subplot using `lifecycle::deprecated()` until the next release of r4ss.

The 4 functions where subplot is used:
```
R\SSplotData.R:
  11  #' @template plotdir
  12: #' @param subplot vector controlling which subplots to create
  13  #' Currently there are only 2 subplots:

R\SSplotSelex.R:
  26  #' Selectivity, 2. Retention, 3. Discard mortality, 4. Keep.
  27: #' @param subplot Vector controlling which subplots to create.
  28  #' Numbering of subplots is as follows,

R\SSplotSpawnrecruit.R:
  6  #' @template replist
  7: #' @param subplot vector of which subplots to show.  1=plot without labels,
  8  #' 2=plot with year labels.

R\SSplotTimeseries.R:
  7  #' @template replist
  8: #' @param subplot number controlling which subplot to create
  9  #' Numbering of subplots is as follows, where the spawning biomass plots
```

iantaylor-NOAA:
The mix of `subplots` and `subplot` was surely just sloppiness on my part. It's grated on me over the years when I've gotten it wrong when using the function, but somehow I never thought to fix it.

The one exception is `SSplotTimeseries()`, where the function can only do one plot at a time. We should probably leave that one as `subplot` and not use a template in the documentation.

k-doering-NOAA:
I know in my own code I have done this sort of thing over and over again, so you are certainly not the only one, @iantaylor-NOAA. I agree that `SSplotTImeseries()` is different and should be left.

I added this change to PR #688 in a new commit.

PR #688, opened 2022-06-15
Issue 311 templates

k-doering-NOAA:
Addresses #311 by adding templates or using existing templates for repeated variables. The full list added:

- [x]         verbose   54
- [x]         plotdir   30
- [x]         pheight   29
- [x]          pwidth   29
- [x]        cex.main   28
- [x]           print   28
- [x]          ptsize   28
- [x]          punits   28
- [x]            plot   27
- [x]          labels   19
- [x]      fleetnames   14
- [x]             dir   13
- [x]             lwd   13
- [x]       legendloc   12
- [x]          fleets   10

I'm still curious if we wanted to do more, such as try to combine some of these templates into one? There are some variables that are always used together, but I can't decide if it is more clear to have them documented in separate templates or in one.


iantaylor-NOAA:
@k-doering-NOAA, this is awesome. Thank you for doing all this cleanup. 

When the templates are grouped within a single file I sometimes have a harder time finding them, so I'm happy with things as they stand now. If in the future having so many individual files becomes a pain for some reason it will be easy to concatenate those files.

k-doering-NOAA:
Thanks @iantaylor-NOAA! I did put 2 of the variables, `pwidth` and `pheight` into the same template, but called it something like `pwidth_pheight.R`, so it should be easy enough to find, I think. But I won't bother putting together any of the other parameters.

@kellijohnson-NOAA , I requested a review from you since @iantaylor-NOAA is out next week, but it can sit if needed until Ian gets back.

k-doering-NOAA:
Resolves #311 , resolves #687

kellijohnson-NOAA:
Regarding the `verbose` argument. The defaults are not all the same in the functions and using the template means that at the moment we don't have guidance/information on what the default value is because it is not always consistent. Should we mention that users can look at the function call or `args()` to see the default value?

R/RebuildPlot.R
```diff
@@ -35,9 +35,8 @@
 #' Default="default".
 #' @param pdf Option to send figures to pdf file instead of plot window in
 #' Rgui. Default=FALSE.
-#' @param pwidth Width of the plot window or PDF file (in inches). Default=7.
-#' @param pheight Height of the plot window or PDF file (in inches). Default=7.
-#' @param lwd Line width for many of the plot elements. Default=2.
+#' @template pwidth_pheight
```
kellijohnson-NOAA, line 38: I see that @k-doering-NOAA and @iantaylor-NOAA started a conversation about the use of a single file for multiple arguments already but from that conversation I did not expect this behaviour. I thought that pwidth_pheight.R would contain
```
@param pwidth, pheight <Description for both parameters>
```
and not two separate calls to `@param`. I think that we should make them two separate files.

R/SS_readctl_3.30.R
```diff
@@ -1767,8 +1766,7 @@ translate_3.30_to_3.24_var_adjust <- function(Variance_adjustment_list = NULL,
 #' @param Q_options The Q options list element in the 3.30
 #'  control file r4ss list output generated from [SS_readctl].
 #' @param Nfleets Number of fleets in the model
-#' @param fleetnames Name of the fleets. Defaults to fleet numbers, in the order
-#' @return A dataframe containing the 3.24 Q setup.
```
kellijohnson-NOAA, line 1771: I think line 1771 was removed by mistake?

R/SSplotDynamicB0.R
```diff
@@ -28,11 +27,10 @@
 #' (single value will apply to both lines).
 #' @param lty Optional vector of line types to be used for the two lines
 #' (single value will apply to both lines).
-#' @param lwd Optional vector of line widths to be used for the two lines
-#' (single value will apply to both lines).
+#' @param lwd Optional vector of line widths to be used for the two lines.
```
kellijohnson-NOAA, line 30: Is there a reason we are not using the template here?

R/SSplotProfile.R
```diff
@@ -36,9 +36,9 @@
 #' Default=TRUE.
 #' @param col Optional vector of colors for each line.
 #' @param pch Optional vector of plot characters for the points.
-#' @param lty Line total for the likelihood components.
+#' @param lty Line type for the likelihood components.
 #' @param lty.total Line type for the total likelihood.
-#' @param lwd Line width for the likelihood components.
+#' @param lwd Line width for the likelihood components.
```
kellijohnson-NOAA, line 41: Use the template here?

man-roxygen/pwidth_pheight.R
```diff
@@ -1,3 +1,5 @@
+#' @param pwidth Default width of plots printed to files in units of
+#' \code{punits}. The default is \code{pwidth=6.5}.
```
kellijohnson-NOAA, line 2: Just curious why you used `\code{<text>}` rather than `punits`? I am not recommending any changes just curious.

man-roxygen/readctl_vars.R
```diff
@@ -1,6 +1,6 @@
 
-#' @param verbose Should there be verbose output while running the file?
-#' Defaults to FALSE.
+#' @param verbose A logical value specifying if output should be printed
```
kellijohnson-NOAA, line 2: This behavior is slightly weird to me to have verbose in another template file when it has its own template file.

k-doering-NOAA: I tried using @template verbose within the template, but it didn't work. I can just remove verbose from this readctl_vars template, though, if that seems more intuitive!

man-roxygen/pwidth_pheight.R
k-doering-NOAA: I just copied and pasted from an old definition and didn't convert it over to markdown syntax. I can make that change, though, or maybe open an new issue to check for instances of \code{} throughout r4ss and change them to backticks (I think that is recommended, correct)?

R/SSplotProfile.R
k-doering-NOAA: Good catch - this one I wasn't sure about because there is also a `@param lwd.total` right below it, and I wasn't sure if the template definition was too general to distinguish between the two?

R/SS_readctl_3.30.R
k-doering-NOAA: Yep, good catch! I'll push up a commit to add it back in.

R/RebuildPlot.R
k-doering-NOAA: Ok, I'll change it back! I made this change before starting that conversations, but was too lazy to change it back.

As a side note, is it possible to have description for both parameters in a single line? I've never seen that!

k-doering-NOAA:
> Regarding the `verbose` argument. The defaults are not all the same in the functions and using the template means that at the moment we don't have guidance/information on what the default value is because it is not always consistent. Should we mention that users can look at the function call or `args()` to see the default value?

I also was wondering about whether mentioning the default within the `@param` tag was good practice or not. The default is always visible within the documentation under usage: 
image 
and maybe it is actually better to not also include it within @param so that they don't actually get out of sync? I'm not sure about this, though and could be convinced otherwise.

man-roxygen/pwidth_pheight.R
k-doering-NOAA: I changed them all to backticks here

man-roxygen/readctl_vars.R
k-doering-NOAA: I removed it from the template

R/RebuildPlot.R
k-doering-NOAA: I separated them into 2 templates.

k-doering-NOAA:
Thanks @kellijohnson-NOAA for taking a look at this so quickly! I just pushed up my changes based on your comments. Let me know if there are any other questions or changes you would like made.

R/SSplotProfile.R
kellijohnson-NOAA: I am fine with leaving it as well as the previous one. There are a few more instances of `@param lwd` though throughout the code. Would you mind making sure we shouldn't change any more of them and then resolving this comment?

kellijohnson-NOAA:
I think it is fine to not have the default to verbose in the definition given that, as you said, users can always look at the function call.

man-roxygen/pheight.R
```diff
@@ -1,5 +1,5 @@
 #' @param pwidth Default width of plots printed to files in units of
```
kellijohnson-NOAA, line 1: Lines 1-2 should be removed, right?

k-doering-NOAA: Of course - sorry for a such a silly mistake!

kellijohnson-NOAA (approved):
Thanks @k-doering-NOAA for all of these commits and the recent changes. I have two more minor things that I think we should change.
* can we change `fleets` to use `@templates fleets` in `get_tuning_table`?
* How come `@param show_in_console` is named intern.R?
Then, we can merge my branch in and merge everything to main.

k-doering-NOAA:
> Thanks @k-doering-NOAA for all of these commits and the recent changes. I have two more minor things that I think we should change.
> 
> * can we change `fleets` to use `@templates fleets` in `get_tuning_table`?
> * How come `@param show_in_console` is named intern.R?
>   Then, we can merge my branch in and merge everything to main.

Thanks @kellijohnson-NOAA  for all the hard work reviewing. Two thoughts:
1. For `fleets` in `get_tuning_table()`, I don't think that it can take the "all" input that fleets can in other functions, which was why I didn't change it.
2. I wonder if show_in_console was a renaming of intern, but it didn't get carried through? I don't see the @template intern being used anywhere currently. I can soft deprecate intern in the 2 functions it is used, rename the template to show_in_console.R, and add `@template show_in_console` to the two functions.

kellijohnson-NOAA:
Thanks @k-doering-NOAA for all of this, great job 🥇 .

R/SSplotProfile.R
k-doering-NOAA: I just took a look - the reason I left them is they all take vectors instead of a single value, which I thought wasn't in line with the template definition. So I think they should be left.

Maybe renaming them something like `lwd_vec` to make them distinct from `lwd` would make sense. I'll open an issue to this effect for us to consider in the future.

k-doering-NOAA:
> Thanks @k-doering-NOAA for all of this, great job 🥇 .

Thanks @kellijohnson-NOAA  for your thorough reviews! I know this took a lot of time, but ended up preventing a lot of bugs for us to find further down the line! Thanks!!!

Once the tests pass, I will go ahead and merge this in.

PR #684, opened 2022-06-14
Test readwrite par

nathanvaughan-NOAA:
This pull request addresses issue #681 and is now passing all tests. Corrections have been made to the read_par function to include 2DAR dev parameters and to the SS_read and SS_write functions to include the ss.par file if it is present.

Issue #681, opened 2022-06-13
SS_readpar_3.30() fails for model with semi-parametric selectivity


iantaylor-NOAA:
In the `test_readwrite_par` branch I added SS_readpar_3.30() to the general test of read/write functions (see https://github.com/r4ss/r4ss/commit/c58ed939056d4e3f06daa8dc3113bbb48ba11499) and see that it's failing (see https://github.com/r4ss/r4ss/runs/6866303767?check_suite_focus=true#step:13:273). 

It's not clear from the github action which of the test models is the source of the failure, but running the function on all those models locally, I see that it's just "Hake_2019_semi-parametric_selex" (files at  https://github.com/nmfs-stock-synthesis/test-models/tree/main/models/Hake_2019_semi-parametric_selex) which fails and the issue is that the semi-parametric (a.k.a. 2D-AR1) selectivity setup creates a `parm_dev` vector in ss.par but unlike other cases there are no parameter lines with a non-zero value in column 9 to trigger filling in the `dev_parm_start` and `dev_parm_end` values that the function is expecting.

@nathanvaughan-NOAA, do you want to take this on as the expert on `SS_readpar_3.30()`, or would you rather that I work on the fix? Either way is fine with me.

Also, @k-doering-NOAA, do you have any tips on figuring out which of the test models causes a failure in the testthat output?

nathanvaughan-NOAA:
Thanks for the detailed investigation @iantaylor-NOAA I can take a crack at it.

k-doering-NOAA:
@iantaylor-NOAA, thanks for pointing this out! I wonder if we could print the name of the model before starting a new test, and that would show up in the console output in github actions, maybe? I suspect I have done as you have in the past, testing locally.

nathanvaughan-NOAA:
@iantaylor-NOAA  and @k-doering-NOAA  I made the correction to the par function but it kept failing. After looking into the tests locally it seems that the issue is in the test-readwrite.R test. I there the SS_read() and SS_write() functions do not read or write the par file. However the final check lapply(files, function(x) expect_true(file.exists(x))) expects the par file to exist. I will add the par file into these functions and see if that clears it up. I'm not sure why this test wasn't failing for every model though??

iantaylor-NOAA:
Thanks @nathanvaughan-NOAA for fixing this so quickly in PR #684.

When modifying the test to add the read/write of the par file I didn't think about the missing link to `SS_read()` and `SS_write()`. 
[...]

PR #689, opened 2022-06-15
Test readwrite par

iantaylor-NOAA:
Completing merge into main that started with #684 from @nathanvaughan-NOAA's fork to this fork.


iantaylor-NOAA:
Tests are now passing and code coverage for read/write par functions jumped up a lot, which was the original goal of the chnage that led to the discovery of the issue in #681.

Thanks @nathanvaughan-NOAA for all the help.

PR #692, opened 2022-06-16
More templates

kellijohnson-NOAA:
I added some more changes to the templating while I was reviewing it. @k-doering-NOAA can you check these then merge them in to your branch?


k-doering-NOAA (approved):
Looks good @kellijohnson-NOAA , thanks for this! I will merge it into the issue_311_templates branch as long as the tests pass.

PR #694, opened 2022-06-17
Document how to customize and host r4ss's html output

### LINKED ISSUE #544: document by which figs and tables get added to html, facilitate posting to github pages.
## Background
GitHub Pages allows anyone with a github account to post any files to the web, including the r4ss html. See https://r4ss.github.io/r4ss/BigSkate/ for example and https://pages.github.com/ for info on the github implementation.

Here's an attempt to document how to do that.

## Current framework
**1. Move existing plots from older models to avoid overwriting** https://github.com/r4ss/r4ss/blob/main/R/SS_plots.R#L408-L437
- `SS_plots()` reads any csv files that start with `plotInfoTable` within the specified `plotdir` (which = `plots` by default).
- If a value in the `StartTime` column of any of those csv files does not exactly equal the start time of the model provided as `replist` to `SS_plots()` then the `plots` directory is renamed to something like `plots_StartTime_Tue_Jun_22_17.34.28_2021` where the value appended in first StartTime value in the first csv file that got read.

**2. Get info on new plots**
- Each function called by `SS_plots()` (e.g. `SSplotBiology()`)returns invisibly a table with details on the plots that got created by that function, e.g. https://github.com/r4ss/r4ss/blob/main/R/SS_plots.R#L497-L505.
- The info provided by each function is `file` (the filename of the png file) and `caption` (which can include html tags and be as complex or as simple as you wish.
- A third variable `category` is added before passing that table back to `SS_output()`, e.g. `bio` for the plots created by `SSplotBiology()`: https://github.com/r4ss/r4ss/blob/main/R/SSplotBiology.R#L1668-L1670. 
- The info for all plots is compiled within `SS_plots()` which then adds two new variables: `png_time` (current time value for `Sys.time()` applied to all plots just made) and `StartTime` (starting time of the model on which the plots were based). The table with the five variables (`file`, `caption`, `category`, `StartTime`, and `png_time` is written to a csv file with name based on the current time (`paste0("plotInfoTable_", format(png_time, "%d-%m-%Y_%H.%M.%OS4"), ".csv")`) where the `%OS4` was just added to use decimal seconds out to 4 decimal places to avoid cases where two files were written within the same second by multiple calls to `SS_plots()`.
   
**3. Create html files**
- The `SS_html` function gets called at the end of `SS_plots() and creates a new tab for each each unique value in `category` in whatever order they appear in any csv file located within the `plotdir` directory (which get read in alphabetical order). If the StartTime values don't match it will stop with a warning which you can override via `multimodel = TRUE`.
- A file `_SS_output.html` gets created with info on the model start time, the SS version, the r4ss installation, and any note provided by the `filenotes` argument to `SS_plot()` (which I only used for the first time yesterday). This file is associated with the category "Home" and I don't know what would happen if any other information were assigned to that category in one of the plotInfoTable csv files.
- A new html file for each category is created with a name set by `paste("_SS_output_", category, ".html", sep = "")` which includes information on all the plots or tables associated with that category in whatever order they appear in the csv files (which are read in alphabetical order). The tabbed output at the top of each html file includes references to all the other files.
- If there are duplicated file names in the csv files (e.g. from multiple calls to `SS_plots()` with different inputs), the one with the more recent `png_time` is used.
- The `_SS_output.html` gets automatically opened in the default browser (unless you run `SS_html(..., openfile = FALSE)` outside of `SS_plots()`.

## adding extra figures to the html view
- Multiple calls to `SS_plots() with different arguments are fine but if you add more panels to a given plot, you might have a lingering "page3" file that when you only want when all the files now fit on pages 1 and 2.
- Any other plots or tables created outside of r4ss that are described in a csv file with matching format should be automatically combined with the ones created by r4ss and if you run `SS_html(..., multimodel = TRUE)` then the StartTime should be ignored.
- Making a csv filename that is alphabetically earlier than the ones created by r4ss should put the extra plots above the other r4ss plots within that category.
- Adding any value for `category` other than the existing defaults should get you a new tab which will probably be in the order it appears among the alphabetically read csv files.

## getting the plots onto github
_to be added_

## future directions to make all this easier within r4ss
_to be added_

### LINKED ISSUE #564: provide information on how to add new plots/tabs to the html viewer for r4ss
@iantaylor-NOAA brought up a while back that this was helpful for Lingcod. Providing some guidance to users on how to do this would be helpful. Perhaps this could be added to the intro vignette, which could be re-rendered.

k-doering-NOAA:
This resolves #544 and resolves #564.

This is still a draft, because it starts a vignette, but I'm not sure if it is headed in the right direction or not. Feedback on the details added to the documentation would be helpful, also. 

I'm adding this as a draft PR to keep it more visible.


k-doering-NOAA:
I won't have time to finish this up, so @iantaylor-NOAA will take it over at some point in the future! Leaving this open as a draft PR just to keep it visible.

k-doering-NOAA:
@iantaylor-NOAA I just discovered this ancient draft PR while looking through my open PRs (something I always forget to do).

Please feel free to delete the branch and close the PR since I think you and @e-perl-NOAA have worked on this and it is likely not necessary anymore!

iantaylor-NOAA:
@k-doering-NOAA, I think your branch remains the best documentation of this topic. It just hasn't risen high enough on the priority list for me to get back to it to finish the work.

k-doering-NOAA:
ok, it's fine to leave it as-is, too!

Issue #698, opened 2022-06-24
write_fwf4 with data.frame having column(s) of factors


yukio-takeuchi:
When I wrote write_fwf4 function, I simplified it to accept columns of numeric( or int) or character.
However, although it is rarely happen with recent version of R with recent SS_read* functions, R may still accidentally identify  one or more of columns of data.frame  as factor.  In fact, I recently met such situation when I manipulated some of data.frame  in a datlist by external R codes.  Within write_fwf4, a piece of code like
```
if (is.factor(x[, I])
  x[,I] <- type.convert(as.character(x[,I]),as.is=TRUE)
```
may avoid error due to column(s) of factor.
I am wondering if this change is necessary or not as no one reported an error of write_fwf4 related to column(s) of factor.

iantaylor-NOAA:
Hi @yukio-takeuchi, this seems like a good idea. Thanks for thinking of it.
Could you be put together a pull request for a change to `write_fwf4()`, or do you want me to just insert the suggested change?

PR #699, opened 2022-06-24
type.convert factor column(s)

yukio-takeuchi:
This change allows to accept data.frame with factor column(s).
 This recolve the problem mentioned in  #698


iantaylor-NOAA:
Thanks @yukio-takeuchi. Tests all passed as expected, so merging.
This closes #698.

Issue #206, opened 2018-01-24
update CRAN version of r4ss


iantaylor-NOAA:
Current version on CRAN is out of date and doesn't work for many 3.30 models.

EDIT 10 May 2022: checklist added from comments below:
- [x] re-read http://r-pkgs.had.co.nz/release.html#release-submission
- [x] update `NEWS.md`
- [x] update `cran-comments.md`
- [x] temporarily add the following to `.Rbuildignore` 
  - [x] `inst/extdata` to avoid NOTE about package size in package check 
  - [x] `tests` because it depends on the example model files in `inst/extdata`
  - [x] `vignettes` because it depends on the example model files in `inst/extdata`
- [x] use `devtools::release()` which asks for confirmation of the following checks
  - [x] `devtools::spell_check()` # will be faster now that `inst/WORDLIST has been created
  - [x] `devtools::check()`
  - [x] `devtools::check_rhub()`
  - [x] `devtools::check_win_devel()` # identified some issues with broken URLs
  - [x] `revdepcheck::revdep_check()` # checks ss3sim and SSMSE
- [x] confirm the submission to CRAN via link in automated email after uploading
- [x] merge `CRAN` branch (if used for submission)
- [x] tag the code with the version number

iantaylor-NOAA:
Hoping to update the CRAN version this week. If it goes well, that might become a more regular occurrence as there have been frequent reports of folks having trouble installing via `devtools::install_github()`.

iantaylor-NOAA:
The following should be done before submission
- [ ] temporarily add `tests` and `inst/extdata` to `.Rbuildignore` to avoid NOTE about package size in package check
- [ ] update `cran-comments.md`
- [ ] re-read http://r-pkgs.had.co.nz/release.html#release-submission
- [ ] use `devtools::release()`

iantaylor-NOAA:
Submitting to CRAN "this week" as I suggested last week didn't work out. Will attempt to make submission next week after consultation with @kellijohnson-NOAA about whether to include changes in Pull Request #341 along with those in Pull Request #340 in the submission.

kellijohnson-NOAA:
@iantaylor-NOAA perhaps we can just use this issue to track submitting to CRAN rather than opening a new one. I tried `rhub::check()` last night but it failed because the tests took too long. I notice now you mentioned adding the tests directory to .Rbuildignore but I am wondering if we should just mark some tests to not run on CRAN rather than ignore the folder?
```
[...]

Issue #306, opened 2019-05-30
Whether new releases of r4ss should maintain backwards compatibility


ChristineStawitz-NOAA:
@CorinneBassin-NOAA and I are working on defining recommended workflows for the NOAA Fisheries Toolbox, and we discussed using github's release feature to simplify branches in code flow. I think breaking backwards compatibility for past SS versions in future releases of r4ss has the potential to simplify the code and improve stability and testability. Although the way the current functions like `read_dat` are modularized is the best way to have a fully-backwards-compatible r4ss, it still seems that many issues are caused when adding functionality for SS 3.30 models breaks functionality for SS 3.24, etc. Furthermore the test suite is much larger when the latest r4ss version needs to work across SS versions.

What I am envisioning is that specific r4ss releases work for some versions but not others; for example, r4ss v x.00 is fully tested and works for SS 2.0 models but not SS 3.30 models, but version x+1.00 of r4ss works for ss 3.30 models only and not SS 2.0 models.

I suggest this because I think it might make your lives easier, but if it seems to make things harder instead (i.e. because you plan to add new features for older SS versions) please let me know. I am hoping to start a conversation, not be prescriptive about changes.

iantaylor-NOAA:
Thanks @ChristineStawitz-NOAA and @CorinneBassin-NOAA for starting this discussion. I agree that maintaining backwards compatibility adds challenges and makes the code more complex and harder to test. However, there are also a variety of areas where it's been really useful.

Two general cases where maintaining compatibility of r4ss with both 3.24 and 3.30 are the following.

**Comparing old and new SS models for the same stock**
The Pacific Fishery Management Council Terms of Reference for groundfish stock assessments (https://www.pcouncil.org/wp-content/uploads/2019/04/Stock_Assessment_ToR_REVISED_2019-20_APR2019_Final-2.pdf) require a description of the bridging from the most recent previous assessment model to the new base model. Plots comparing biomass time series using the `SSplotComparisons` function has been a useful part of this description. The 2017 assessments were conducted in SS version 3.30, but many assessments conducted in 2015 and before have not yet been replaced by full assessments, and would benefit from r4ss compatibility when that happens. I suspect a similar process is in place for other groups who use SS. However, even without compatibility, it's not that hard to extract a biomass time series from an old version of a Report file and add it to the input to `SSplotComparisons`. 

[...]

PR #700, opened 2022-06-24
add smaller test model, refactor tests

iantaylor-NOAA:
This Pull Request includes the following:
- adds a new "simple_small" model
  - based on "simple"
  - has fewer years (main time series is 2011-2022)
  - minor changes to model configuration for simplicity or to implement more common settings
  - files are about 2 MB which should be small enough to keep the package within the 5 MB limit on CRAN
  - changes from "simple" made via script attached to this comment https://github.com/r4ss/r4ss/issues/206#issuecomment-1157078047
- refactors the tests to put all tests associated with 3.24 and earlier 3.30 versions into a single `test-older_ss3.R` file which can be excluded from future CRAN submissions along with those model files
- other tests should work on CRAN with the exception of those that require the executable which will be skipped

I think this draft pull request will trigger a code coverage check which will help me see if I've missed some important pieces.
Feel free to comment on any of this.


k-doering-NOAA:
@iantaylor-NOAA, good call on making a smaller model. I think we should go ahead and replace any of the simple 3.30 models with this model!

I'm curious, do you think we should retain a 3.24 model for testing (even if that testing is skipped on CRAN)? And if so, should we try to make it smaller like this one?

k-doering-NOAA:
Also, I took a brief glance through the diffs and the changes make sense to me. I haven't scrutinized everything line-by-line, though, as I figure I can do that when this is no longer a draft. Happy to take a closer look if you would like, though!

iantaylor-NOAA:
@k-doering-NOAA, good questions. I thought about removing the old models, but it doesn't make sense to maintain the read/write functions for 3.24 models if we don't maintain the tests that check those functions.

We had a discussion about backwards compatibility in #306 and I think this PR is in line with the views shared there: maintaining tools for accessing and understanding models associated with old versions of SS3 (`SS_output()`, `SS_plots()`, and the read/write functions for the input files), but no longer testing things like the profile, retrospective, or jitter functions for old versions.

However, it surely doesn't make sense to include the files associated with those older version as part of the package that all users get when installing r4ss from github. Would it be possible to create a new repo with the simple models from the older versions of SS3 and revise the r4ss-extra-tests workflow to copy them for testing purposes? That would also save us from removing those files from future CRAN submissions.

k-doering-NOAA:
@iantaylor-NOAA , should the simple_3.30.xx models in inst/extdata be removed before merging in, since the smaller test can serve as a replacement?

tests/testthat/test-control.R
```diff
@@ -13,242 +13,108 @@ file.copy(example_path, tmp_path, recursive = TRUE)
 on.exit(unlink(tmp_path, recursive = TRUE))
 
 # paths to the model files
-sim_3.30.13 <- file.path(tmp_path, "extdata", "simple_3.30.13")
-sim_3.24 <- file.path(tmp_path, "extdata", "simple_3.24")
+sim_small <- file.path(tmp_path, "extdata", "simple_small")
 
-test_that("SS_readctl and SS_writectl works for 3.30.13", {
```
k-doering-NOAA, line 19: I think the other 3.30.xx version tests can be removed in this PR.

tests/testthat/test_Sensi_plot.R
```diff
@@ -59,7 +59,7 @@ test_that("SS_Sensi_plot runs", {
   SS_Sensi_plot(
     model.summaries = model.summaries,
     dir = temp_path,
-    current.year = 2001,
+    current.year = 2022,
```
k-doering-NOAA, line 62: Good idea to make the last model year 2022 instead of 2001!

k-doering-NOAA (changes requested):
Hi @iantaylor-NOAA , it looks good. I asked a few questions for you to consider. Once done with changes, I will change the workflows to reference `@main` instead .

tests/testthat/test-control.R
```diff
@@ -13,242 +13,108 @@ file.copy(example_path, tmp_path, recursive = TRUE)
...
-    ctllist = ctl_3.24,
-    version = "3.24",
-    verbose = FALSE,
-    overwrite = TRUE,
-    outfile = file.path(sim_3.24, "testctl.ss")
-  )
-  expect_true(file.exists(file.path(sim_3.24, "testctl.ss")))
-})
-
-test_that("SS_readctl and SS_writectl works for 3.24 when datlist = FALSE", {
```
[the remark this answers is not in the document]
iantaylor-NOAA: Tests removed as discussed in #702

iantaylor-NOAA:
After way to many typos and omissions, I think I've got these changes cleaned up and the suggestions from @k-doering-NOAA implemented and tests passing, so will merge this branch.

There remains cleanup to deprecate and then remove the 3.24-related functions in #702 but that can be separate and happen later. Also, we may still want to move the 3.24 model and earlier 3.30 models used for testing `SS_output()` into a different repo and create a new github action to use it for testing.

Code coverage has decreased because the read/write functions for 3.24 models aren't being tested, and some of the elements of `SS_output()` that are specific to earlier 3.30 versions are also being skipped. But I'm OK with all those reductions.

k-doering-NOAA:
Thanks @iantaylor-NOAA ! I can work on deprecating the 3.24 functions.

Issue #703, opened 2022-07-01
refine phase plot


iantaylor-NOAA:
A data and stock assessment open discussion (I think on May 6) focused on phase plots identified a few potential improvements.
These changes have been attempted in the [phase_plot_changes](https://github.com/r4ss/r4ss/tree/phase_plot_changes) branch, including
- add a reference point line at the SPR target when `SPR_report_basis` has SPRratio equal to raw SPR (and thus the target doesn't correspond a ratio of 1.0)
- add labels to point at each decade in addition to the status-quo labels on the start and end year (this leads to some overlapping labels which I can't be bothered to sort out, but am open to ideas)
- expand the caption to note the relationship between color and time period

I made these changes a while ago and then forgot about them so I'm not sure what further work was needed before merging this into the main branch. Maybe I just wanted some feedback on whether this looks OK or if folks have other suggested changes. Please comment below if you have suggestions.

**Example of status-quo phase plot:**
![image](https://user-images.githubusercontent.com/4992918/176941193-fd5ba2e2-b4ae-4eb8-a4b0-873a99ff2a1e.png)
**Modified phase plot:**
![image](https://user-images.githubusercontent.com/4992918/176941111-d72829b7-90c2-4a52-8fd7-8f2bdae3db21.png)

kellijohnson-NOAA:
@iantaylor-NOAA overplotting of labels is easily dealt with using [ggrepel](https://github.com/slowkow/ggrepel) but I am guessing that change isn't what you are looking for 😉. I tried [`plotrix::spread.labels()`](https://www.rdocumentation.org/packages/plotrix/versions/3.7-2/topics/spread.labels) as suggested on [stack exchange](https://stats.stackexchange.com/questions/16057/how-do-i-avoid-overlapping-labels-in-an-r-plot) and found it did not quite work that great, i.e., some labels are outside of the figure.
![plot](https://user-images.githubusercontent.com/4108564/176947097-92be6181-3eb5-496a-be25-b59e8af575fe.png)

iantaylor-NOAA:
@kellijohnson-NOAA, thanks for trying out a solution even though it didn't work that well. I look forward to {ggrepel} in future packages, but for now I think living with the overlap is fine. 

In general there will be little mystery about what any unreadable overlapping labels say as it will typically just be all the decades before relative fishing intensity went up relative spawning output went down.

iantaylor-NOAA:
Forgot to close this issue after merging #704.

PR #704, opened 2022-07-06
Phase plot changes

iantaylor-NOAA:
This branch includes changes to the phase plot discussed in #703. 
I don't think these need a formal review but I'll let the github actions complete the testing before merging.


R/SSplotSPR.R
```diff
@@ -28,6 +28,9 @@
 #' sprtarg input here will cause no horizontal line to be plotted.
 #' @param btarg target depletion to be used in plots showing depletion. May be
 #' omitted by setting to NA. "default" chooses based on model output.
+#' @param minbthresh minimum biomass thresholdto be used in plots
```
k-doering-NOAA, line 31: missing space between "threshold to"

R/SSplotSPR.R
```diff
@@ -45,7 +48,7 @@ SSplotSPR <-
            uncertainty = TRUE,
            subplots = 1:4, forecastplot = FALSE,
            col1 = "black", col2 = "blue", col3 = "green3", col4 = "red",
-           sprtarg = "default", btarg = "default",
+           sprtarg = "default", btarg = "default", minbthresh = "default",
```
k-doering-NOAA, line 51: I wonder if this should instead be `minbthresh = replist[["minbthresh"]]`, as I think making defaults directly assigned when they are simple enough is supposed to be a good practice? Then there is no need to explain what default means in the documentation.

R/SSplotSPR.R
```diff
@@ -427,28 +446,59 @@ SSplotSPR <-
...
+      # if SPRratio denominator is not SPR target,
+      # then add a line at sprtarg
+      if (sprtarg > 0 & SPRratioLabel == "1-SPR") {
+        abline(
+          h = 1 - sprtarg,
           lty = 2,
           col = rgb(0, 0, 0, 0.4)
         )
+        # caption <- paste0(
+        #   caption,
```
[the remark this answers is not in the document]
iantaylor-NOAA: Looks like it was unfinished changes. Needed revision and moving around.

R/SSplotSPR.R
iantaylor-NOAA: I will open a separate issue to apply this to other functions.

R/SSplotSPR.R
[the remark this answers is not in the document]
iantaylor-NOAA: Done.

PR #718, opened 2022-07-20
Deprecate 3.24 functions

### LINKED ISSUE #702: deprecate functions relating to 3.24 and older models
@iantaylor-NOAA and I had a discussion today about whether it makes sense to continue maintaining code that is specifically for 3.24 and older SS3 models. We think it makes sense to deprecate them if they are not getting a lot of use because there is some cost to maintaining them (or in some cases, we are not testing the functions at all, which means they are not reliable for r4ss users).

The list of functions we are thinking of deprecating are:
- `SS_readdat_2.00()`
- `SS_readdat_3.00()`
- `SS_readdat_3.24()`
- `SS_readctl_3.24()`
- `SS_readpar_3.24()`
- `SS_writedat_3.24()`
- `SS_writepar_3.24()`
- `SS_writectl_3.24()`

It could also be possible to greatly simplify other functions that currently take the `version` argument and have code for version other than 3.30. I'm not sure if we would want to deprecate the version function entirely or just warn if it is not a valid value? 

We are wondering if anyone sees large value in these functions or would be concerned if they were deprecated? It would be great to hear other's opinions.

@nathanvaughan-NOAA , I was wondering if you had thoughts, as an author and user of some of the 3.24 read/write functions.

k-doering-NOAA:
Resolves #702 by deprecating read and write functions related to 3.24.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: r4ss
 Type: Package
 Title: R Code for Stock Synthesis
-Version: 1.45.2
+Version: 1.45.3
 Authors@R: c(person(given = "Ian G.", family = "Taylor",
   email = "Ian.Taylor@noaa.gov", role = c("aut","cre")),
   person(given = "Ian J.", family = "Stewart", role = "aut"),


R/SS_changepars.R
```diff
@@ -331,7 +331,7 @@ SS_changepars <-
     if (verbose) {
       message(
         "Wrote new file to ", newctlfile, " with the following changes:\n",
-        paste0(capture.output(results), collapse = "\n")
+        paste0(utils::capture.output(results), collapse = "\n")
```
iantaylor-NOAA, line 334: I like the practice of adding the package name to functions like this in general because it makes clear where the function is coming from and easier to figure out elements of the required packages are being used. However, in this case {utils} is part of the base R installation, so what does it add? Certainly there's no harm in adding this, I'm just curious.

iantaylor-NOAA (approved):
Looks good. Thank you @k-doering-NOAA.
This resolves issue #702.

Can you also push a change to DESCRIPTION to bump the version number to 1.45.3 so folks start getting the message right away?

R/SS_changepars.R
k-doering-NOAA: There was a warning about this in r cmd check, which is the only reason I added it!

k-doering-NOAA:
Thanks for reminding me to bump the version in the DESCRIPTION file. I'll let the github actions finish and then will merge this in!

R/SS_changepars.R
iantaylor-NOAA: Makes sense. Thank you for making the warning go away.

PR #721, opened 2022-07-21
Create a function download_models to download ss3 test models

### LINKED ISSUE #574: Create function to download "models" directory and provide documentation

k-doering-NOAA:
Resolves #574  by creating a `download_models()` function that will download the models subfolder from [the nmfs-stock-synthesis/test-models repository](https://github.com/nmfs-stock-synthesis/test-models). There is an option to specify which branch to download and which directory to write to.

I requested @kellijohnson-NOAA to review as I think this was her idea, but feel free to reassign to @iantaylor-NOAA instead if you are too busy.


R/download_models.R
```diff
@@ -0,0 +1,50 @@
...
+#' @references [nmfs-stock-synthesis/test-models repository](https://github.com/nmfs-stock-synthesis/test-models#test-models)
+#' @export
+download_models <- function(dir = file.path("inst", "extdata"), 
+  branch = "main", overwrite = FALSE) {
+  # checks
+  if (!dir.exists(dir)) {
+    stop("dir ", dir, " does not exist")
+  }
+  zip_file_path <- file.path(dir, "test-models.zip")
+  result <- tryCatch(utils::download.file(
```
kellijohnson-NOAA, line 26: I would prefer to change 26-33 to
```
  result <- tryCatch(
    utils::download.file(
      url = paste0(
        "https://github.com/nmfs-stock-synthesis/test-models/archive/",
        branch,
        ".zip"
        ),
      destfile = zip_file_path
    ),
    warning = function(e) {
      stop(
        "The test-models zip file could not be downloaded.",
        " Does the branch (", branch, ") exist?",
        call. = FALSE
      )
    }
  )
```
where the indents make it more clear what `tryCatch` is doing and the behavior of the warning is linked to the `stop()`.

R/download_models.R
```diff
@@ -0,0 +1,50 @@
...
+      url = paste0("https://github.com/nmfs-stock-synthesis/test-models/archive/",
+        branch, ".zip"),
+      destfile = zip_file_path),
+       error = function(e) e)
+  if("simpleError" %in% class(result)) {
+    stop("The test models could not be downloaded. Does the branch exist?")
+  }
+  list_files <- utils::unzip(list = TRUE, zipfile = zip_file_path)
+  save_files <- list_files[grep("/models/", list_files$Name, fixed = TRUE), ]
```
kellijohnson-NOAA, line 35: Will the forward slashes around models always be forward regardless of the operating system?

R/download_models.R
```diff
@@ -0,0 +1,50 @@
...
+  utils::unzip(zipfile = zip_file_path, files = save_files[["Name"]], 
+    exdir = dir)
+  if (dir.exists(file.path(dir, "models")) & overwrite == FALSE) {
+    warning("The model directory ", file.path(dir, "models"),  " already exists ",
+      "\nand overwrite is FALSE, so no new files will be written." )
```
kellijohnson-NOAA, line 40: Change `, so no` to `. So, no`. Also, remove space before end bracket.

R/download_models.R
```diff
@@ -0,0 +1,50 @@
...
+  }
+  dir.create(file.path(dir, "models"), showWarnings = FALSE)
+  copy_status <- file.copy(from = file.path(dir, paste0("test-models-", branch), "models"), 
+            to = file.path(dir), recursive = TRUE, overwrite = overwrite)
+  # clean up
+  unlink(zip_file_path)
+  unlink(file.path(dir, paste0("test-models-", branch)),
+    recursive = TRUE)
+  invisible(copy_status)
+  }
```
kellijohnson-NOAA, line 50: add return carriage at end of file.

R/download_models.R
```diff
@@ -0,0 +1,50 @@
+#' Download SS3 test models
+#'  
+#' Function to download the models folder within the 
+#' nmfs-stock-synthesis/test-models repsitory
```
kellijohnson-NOAA, line 4: Add full stop at EOL.

tests/testthat/test-download_models.R
```diff
@@ -0,0 +1,55 @@
...
+})
+
+test_that("download_models with different dir works", {
+    download_models(dir = save_models_dir)
+    expect_true(file.exists(file.path(save_models_dir, "models")))
+    expect_true(file.exists(file.path(save_models_dir, "models", "Simple")))
+    expect_true(file.exists(file.path(save_models_dir, "models", "Simple", "ss.par")))
+})
+
+# test_that("download models works with different branch", {
```
kellijohnson-NOAA, line 32: Should we delete this rather than leave in an unused test? I am guessing it is commented out because branches other than main will be transient?

tests/testthat/test-download_models.R
```diff
@@ -0,0 +1,55 @@
...
+#     expect_true(file.exists(file.path(temp_dir_branch, "models", "Simple", "ss.par")))
+# })
+
+test_that("download_models() fails when the branch doesn't exist", {
+    temp_dir_branch_dne <- file.path(save_models_dir, "diff_branch_dne")
+    dir.create(temp_dir_branch_dne)
+    expect_error(
+      download_models(dir = temp_dir_branch_dne, 
+                      branch = "not_existing_branch"), 
+      "The test models could not be downloaded")
```
kellijohnson-NOAA, line 47: I prefer to not have so much text in the checking of an error message because the error message can change with development, in fact I think I suggested a change to the message already. 🤦

kellijohnson-NOAA (approved):
@k-doering-NOAA thanks for writing this function. It works great. I tested it out on my local machine but I did not run the tests. GitHub actions is enough for me there 😉 

Feel free to ignore all of the nit picky comments. But, there is one substantial design-choice that I am not sure that I agree with. Instead of just stopping if `dir` doesn't exist cannot we create `dir` recursively at the start of the function?

k-doering-NOAA:
Thanks @kellijohnson-NOAA for the comments! I'll go over them later today or early next week.

I was just reading the ropen sci blog post and wanted to drop a link on this thread; when I was developing the testthat tests for `download_models()`, I tried to use `tempdir()` directly and it was a disaster! Apparently what you want to do is create a subdirectory within tempdir to do your testing, which is the approach I arrived at by trial and error, so it is good to see this advice reflected elsewhere as the best practice to use: https://ropensci.org/blog/2022/01/31/package-build-failures/#whacky-use-of-temporary-directories

R/download_models.R
```diff
@@ -0,0 +1,50 @@
...
+#'   model_name <- list.files("models") # see the model names
+#'   # remove files
+#'   unlink(file.path("models"), recursive = TRUE)
+#' @author Kathryn Doering
+#' @references [nmfs-stock-synthesis/test-models repository](https://github.com/nmfs-stock-synthesis/test-models#test-models)
+#' @export
+download_models <- function(dir = file.path("inst", "extdata"), 
+  branch = "main", overwrite = FALSE) {
+  # checks
+  if (!dir.exists(dir)) {
```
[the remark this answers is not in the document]
k-doering-NOAA: Good point! Honestly, I was just lazy :P I can add that in.

tests/testthat/test-download_models.R
k-doering-NOAA: Yes, all other branches are feature branches, so I'll just delete the test.

R/download_models.R
```diff
@@ -0,0 +1,50 @@
+#' Download SS3 test models
+#'  
+#' Function to download the models folder within the 
+#' nmfs-stock-synthesis/test-models repsitory
+#' @param dir The folder where the models subfolder should be written to.
+#' @param branch The name of the github branch to download.
```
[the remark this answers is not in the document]
k-doering-NOAA: Thank you for all the specific wording changes! putting them in a code chuck also makes it easy to copy/paste. I'm going to follow that practice when I suggest specific code or wording from now on!

tests/testthat/test-download_models.R
k-doering-NOAA: I'll change it to be less text. I guess the reason why I like to put at least some text is to make sure it captures the error message I'm expecting, but I can make it capture fewer words.

R/download_models.R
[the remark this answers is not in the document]
k-doering-NOAA: I'm glad you pushed me to do this, because it took one line of code that I already knew how to write 😆

R/download_models.R
```diff
@@ -0,0 +1,50 @@
...
+#' @examples
+#'   download_models(dir = getwd())
+#'   model_name <- list.files("models") # see the model names
+#'   # remove files
+#'   unlink(file.path("models"), recursive = TRUE)
+#' @author Kathryn Doering
+#' @references [nmfs-stock-synthesis/test-models repository](https://github.com/nmfs-stock-synthesis/test-models#test-models)
+#' @export
+download_models <- function(dir = file.path("inst", "extdata"), 
+  branch = "main", overwrite = FALSE) {
```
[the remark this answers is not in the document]
k-doering-NOAA: thanks, I realized I needed to run styler, my bad.

R/download_models.R
k-doering-NOAA: So, I tried this out, and when using warning= instead of error= it creates a situation where the zip file connection is left open and so isn't properly deleted (because it is being accessed by a different program). This then means the tests can't be fully cleaned up. I found changing to `error=` prevents this unintended consequence from happening; Do you think that is an acceptable compromise?

kellijohnson-NOAA: I think that should be fine.

k-doering-NOAA: Thanks, I will push my changes shortly so you can see them!

kellijohnson-NOAA:
@k-doering-NOAA sorry for this late comment but did you look at {usethis} for working with zip files, see [here](https://github.com/r-lib/usethis/blob/HEAD/R/course.R)?

k-doering-NOAA:
Aww, I didn't know about this! I actually looked at usethis, because there is a function that can download a file from github, but I didn't know about usethis::use_zip ! 

Do you think I should modify the PR? The plus of what I wrote is that we don't need to add usethis as a dependency to r4ss, but I can change the code if you would prefer!

kellijohnson-NOAA:
I think @iantaylor-NOAA should decide if we want another dependency or not. But, I like that usethis in theory provides the checks rather than us having to write them and it ensures that the correct files are closed. Leaving the code is fine. The benefit of using {usethis} is that it is easier for people to copy the single function into their own code and use it or for us to reuse it later. Not sure if that is going to be a thing here though.

iantaylor-NOAA:
I think making use of the {usethis} function makes sense if it's not too much work to revise all the work that @k-doering-NOAA already put into this. There's nothing about downloading zip files that will ever be unique to {r4ss}, so I'm happy to avoid maintaining custom code when we can share what others are doing and focus on the unique parts of pointing to the model files that we want to download.

I've also come to see a difference between depending on less common packages with questionable maintenance (like {gtools} as discussed in https://github.com/ss3sim/ss3sim/issues/383) vs those that have the full weight of the paid RStudio (now Posit: https://posit.co/) staff maintaining them. The {usethis} package seems genuinely useful so nudging {r4ss} users who don't yet have it installed to do so might improve their lives in other ways.

k-doering-NOAA:
I looked a little more into `usethis::use_zip()` and think it is more for interactive use, as it always asks the user if they want to delete the .zip file or not (I wasn't sure how to suppress this). Given this, and the fact that we will still need some code to delete all files besides the `models` subfolder, I think leaving this PR as-is makes the most sense. This can be merged in if Ian and/or Kelli agree!

kellijohnson-NOAA:
Thanks for checking @k-doering-NOAA into {usethis}. I think that instead of suppressing the warning we could ask {usethis} in the future to remove the interactive nature of the function if we ever did want to change. But, I agree that the code that you wrote is a better solution for now given the other one-off file management that must occur. Sorry for the 🐰 hole.

k-doering-NOAA:
Good idea re: a feature request! Also, the function looks useful for instructing folks on how to download a github repo without needing to clone, so it was definitely worth the time looking into it!

Issue #460, opened 2020-12-29
create a single function for running SS models for use in r4ss, SSMSE, and ss3sim


iantaylor-NOAA:
While discussing issues #430 and #456, @k-doering-NOAA and I realized that not only are there numerous inconsistent approaches within r4ss for running SS models, there are redundancies across packages with 
[runSS](https://github.com/nmfs-fish-tools/SSMSE/blob/main/R/runSS.R) in SSMSE, [run_ss3model](https://github.com/ss3sim/ss3sim/blob/master/R/run_ss3model.r) in ss3sim, and [run_SS_models](https://github.com/r4ss/r4ss/blob/development/R/run_SS_models.R) in r4ss.

It would be great to replace all these with a single function which could do the following:
- deal with exe in path, local directory, or directory provided (as noted in #429)
- add "./" for Mac and Linux systems but only when the executable is not in the path
- display the console output or save it to a file depending on what the user wants
- use `system2()` (as noted in #456)
- be useful for all the functions in r4ss that run SS (as noted in #430)
- be useful for all the needs of SSMSE and ss3sim
- have a short name (I like "runSS" best) and allow users to simply call `runSS()` to run SS within the working directory without additional input
- any other features that anyone adds to this issue

iantaylor-NOAA:
@k-doering-NOAA, thanks for the cleanup to use `show_in_console` throughout the functions that call the executable in commit https://github.com/r4ss/r4ss/commit/0c55d6daf435bc0b201e3b2cc3b45020f663ceb9.
Do you think it's still worth pursuing the shift proposed in this issue to using `run_SS_models()` (or equivalent) in all of those functions?

I think the first three bullets in the original post are already included in that function. We would still need to change to `system2()`.
I like the idea of shortening the name as suggested by @kellijohnson-NOAA, maybe down to just `r4ss::run()`?

iantaylor-NOAA:
Update: I made some progress on this in the [system2_issue456](https://github.com/r4ss/r4ss/tree/system2_issue456) branch (where I was associating the commits with related issue #456). There's still more work to do, especially testing all the changes I've made so far, but it's satisfying to see how much duplicate code can be removed (e.g. about 1/3 of SS_doRetro() [went away](https://github.com/r4ss/r4ss/commit/89bb47ead369d13a7003d03f10c49e2f815ddf65#diff-c3483e7a05e48b3a835217b3f86415f787d27d8b58a2b4a267641e2ea4b6c3c7)).

[...]

Issue #429, opened 2020-10-07
support workflow of having SS executable in the path rather than directory with model files


iantaylor-NOAA:
The SS User Manual notes workflow options under "Running SS", "Command Line Interface": "On all operating systems, a copy of the SS executable can either be located in the same directory as the model input files or in a central location and referenced either by adding it to the PATH or by a script files."

However, the `SSRunJitter()` and `SS_doRetro()` functions (and probably others too) only work with the executable in the same directory as the model input files.

It should be possible to modify the `check_model()` function to look for availability of `ss` in the path, such as by using a command like `system("where ss")` and then change all functions like `SS_doRetro()` to use `check_model()` (it is used by `SSRunJitter()` already).

k-doering-NOAA:
I think from R, you can use
```r
bin_path <- Sys.which(bin_name)[[1]]
system(bin_path)
```
to find and run an executable called bin_name (for example, "ss") in the user's path.

Also, if the user is able to supply a relative (to the working directory) or absolute path to the executable, it should be possible to just include this path in the system call to run ss. For example, if the relative path to the wd is "./exe/ss.exe" running 
```
system("./exe/ss")
```
 should work.

These are a bit specific, but here are the [utility functions for running ss in SSMSE](https://github.com/nmfs-fish-tools/SSMSE/blob/master/R/runSS.R) (modified from the functions in ss3sim!) for more inspiration.

k-doering-NOAA:
@iantaylor-NOAA , I left this open since I see you referenced it recently and gave it an `Epic` label! 🙌🏻 

Feel free to close if you don't need it, though.

kellijohnson-NOAA:
@iantaylor-NOAA I haven't taken the time to work all the way through a downstream function that uses `check_exe()` but what happens if I have an exe in the directory and in my path and I do NOT want to use the one in the path?

iantaylor-NOAA:
Good catch @kellijohnson-NOAA. I have the logic backwards in the function.

"Windows always looks in the current directory first before searching the path" (from [here](https://stackoverflow.com/questions/23455637/in-windows-cmd-how-do-i-run-an-executable-in-the-current-directory-instead-of)).

[...]

PR #722, opened 2022-07-21
revamp functions that run executable and associated cleanup

iantaylor-NOAA:
A bunch of major changes included here as discussed in #460: 
* Major revamp of all functions that run the Stock Synthesis executable
* `check_model()` is now `check_exe()` and supports having an executable in the path (#429). Here's the new "description" section in the function documentation: 
    > First searches path for the specified executable name and returns the location of that file if found. If not found in path it then searches the specified directory. Linux systems have an existing executable utility `/usr/sbin/ss` in the path. If `exe = "ss"` and this file is found by this function, it will be ignored based on the smaller file size relative to the SS3 executable. Linux users who want to use the workflow of having an SS3 executable in their path should name the file something else, such as `ss3` or `ss_linux`. 
* `run_SS_models()` is now `run()` and has easier defaults and better
    support for having the Stock Synthesis executable in your path (by relying on `check_exe()`)
* Functions have more consistent inputs (e.g. always `dir` instead of `dir`, `mydir`, `File`, or `masterdir`, and now defaulting to current working directory)
* Input `show_in_console = FALSE` will pipe output to a text file keeping R console cleaner while models run
* Adds dependency on 'tidyr'
* Removes never-completed function `SSbootstrap()`

Here are a couple design choices that I would appreciate additional input on before removing "draft" from this Pull Request:
1. The functions that run the executable use `...` to pass additional arguments to `run()`, such as `extras`, `show_in_console`, and `skipfinished` rather than add a bunch of new inputs to each of those functions (previously not all these options were available or they were controlled by inconsistently named inputs). Benefit is that the function isn't cluttered with as many inputs, but the drawbacks are that these options are less visible and error message associated with incorrect inputs are further removed from the function the user called as in the message where the user calls the jitter function and gets an error from the run() function:
   ```
   > SS_RunJitter(dir = "inst/extdata/simple_small", Njitter = 1, nonsense = "blah")
   Error in run(dir = dir, exe = exe, verbose = verbose, ...) : 
     unused argument (nonsense = "blah")
   ```

2. Can we rename the functions that call the executable as follows to avoid a bunch of inconsistent capitalization? I didn't do this in the pull request because it makes reviewing changes harder (I probably should have held off on renaming run_SS_models() to run() for that reason as well).

  old | new
  -- | --
  SS_RunJitter() | jitter()
  SS_doRetro() | retro()
  SS_profile() | profile()
  SS_tune_comps() | tune()


R/SS_doRetro.R
```diff
@@ -60,184 +53,76 @@
...
-
-  if (!file.exists(newdir)) dir.create(newdir)
-
   subdirnames <- paste0(subdirstart, years)
 
+  # check for executable
+  check_exe(exe = exe, dir = olddir, verbose = verbose)
+
+  # loop over retrospective years
   for (iyr in 1:length(years)) {
```
k-doering-NOAA, line 89: I suggest changing `1:length(years)` to `seq_along(years)`

R/SS_output.R
```diff
@@ -573,21 +564,29 @@ SS_output <-
...
+        logfile <- tidyr::separate(as.data.frame(logfile),
+          col = 1,
+          into = c("File", "Size"),
+          sep = " = "
+        )
+        names(logfile) <- c("TempFile", "Size")
+        logfile[["Size"]] <- as.numeric(logfile[["Size"]])
+        maxtemp <- max(logfile[["Size"]])
+        if (maxtemp == 0) {
+          if (verbose) {
```
k-doering-NOAA, line 580: nitpick: I think `if (verbose)` could go before `if (maxtemp = 0)` so that it could apply to the `else()` statement, too, instead of having another if (verbose)` statement within the else.

R/SS_tune_comps.R
```diff
@@ -193,8 +185,8 @@
 SS_tune_comps <- function(replist = NULL, fleets = "all",
                           option = c("Francis", "MI", "none", "DM"),
                           digits = 6, write = TRUE, niters_tuning = 0,
-                          init_run = FALSE, dir = getwd(), model = "ss",
```
k-doering-NOAA, line 196: I wonder if `model` should be formally deprecated using lifecycle::deprecated, instead of just changing the name to exe?

R/check_exe.R
```diff
@@ -0,0 +1,83 @@
+#' Find location of executable within path or specified directory
+#'
+#' Check that the executable name provided in `model`,
```
k-doering-NOAA, line 3: I think this documentation may need to be updated, as I think instances of `model` have been changed to `exe` in this PR?

May also want to make it clear that this is an "ss3 executable" and not just any old executable.

R/copy_SS_inputs.R
```diff
@@ -135,12 +135,21 @@ copy_SS_inputs <- function(dir.old = NULL,
...
-      exefiles <- list.files(dir.exe, pattern = "^[^.]+$")
+      # exefiles <- list.files(dir.exe, pattern = "^[^.]+$")
+
+      # directories are accidentally being included using line above,
+      # so trying a different way to exclude them
+      exefiles <- dir(dir.exe)[!file.info(dir(dir.exe,
+        full.names = TRUE
+      ))[["isdir"]]]
+      exefiles <- grep(pattern = "^[^.]+$", x = exefiles, value = TRUE)
       message("Unix binaries are: ", paste0(exefiles, collapse = ", "))
```
k-doering-NOAA, line 153: maybe wrap this message in `if (verbose)`?

R/getADMBHessian.R
```diff
@@ -3,17 +3,32 @@
...
 #' @author Cole Monnahan
 #' @export
 #' @seealso [read.admbFit()], [NegLogInt_Fn()]
 #' @note Explanation of the methods (in PDF form):
 #' <https://github.com/admb-project/admb-examples/blob/master/admb-tricks/covariance-calculations/ADMB_Covariance_Calculations.pdf>
-getADMBHessian <- function(File = getwd(), FileName = "admodel.hes") {
-  filename <- file(file.path(File, FileName), "rb")
+getADMBHessian <- function(dir = getwd(),
+                           File = lifecycle::deprecated(),
+                           FileName = "admodel.hes") {
```
k-doering-NOAA, line 18: A bigger picture thing I've wondered about with r4ss is whethere functions should take a dir and a FileName separately, or fi they should just take a filepath (i.e., `filepath = file.path(dir, FileName)`. Do you have thoughts on this? I think there is a mix of approaches throughout r4ss.

R/run.R
```diff
@@ -0,0 +1,150 @@
+#' Run multiple Stock Synthesis models
+#'
+#' Loops over a vector of directories and iteratively runs SS in each one
```
k-doering-NOAA, line 3: Change SS to SS3

R/run.R
```diff
@@ -0,0 +1,150 @@
...
+#' @template exe
+#' @template extras
+#' @param skipfinished Skip any folders that already contain a Report.sso file.
+#' This can be helpful if the function is interrupted.
+#' @template show_in_console
+#' @param console_output_file File to store console output (if
+#' show_in_console = FALSE)
+#' @template verbose
+#'
+#' @return Returns table showing which directories had model run and which
```
k-doering-NOAA, line 16: model -> models (grammar)

R/run.R
```diff
@@ -0,0 +1,150 @@
...
+#' had errors like missing executable or Report.sso already present.
+#' The three possible messages are "model run failed", "ran model", and
+#' "unknown run status".
+#'
+#' @description First searches path for the specified executable name
```
k-doering-NOAA, line 21: I think this paragraph could use some grammar/clarity edit. e.g., 

```
#' @description The `run` function first searches the PATH for the 
#' specified SS3 executable name
#' and returns the file's location if found. If the SS3 executable is not found in the PATH, 
#' then it is assumed the that executable exists in the specified directory. 
#' Linux systems have an
#' existing executable utility `/usr/sbin/ss` in the path. If `exe =
#' "ss"` and this file is found by this function, it will be ignored
#' based on the smaller file size relative to the SS3 executable. Linux
#' users who want to use the workflow of having SS3 in
#' their PATH should name the SS3 file something besides `ss`, such as `ss3` or
#' `ss_linux`. 

```

(note that this text needs to be re-wrapped)

man-roxygen/exe.R
```diff
@@ -0,0 +1,2 @@
+#' @param exe Executable name, optionally including the path to the
```
k-doering-NOAA, line 1: maybe capitalize `PATH` to distinguish it better just any folder path. This sentence is a bit confusing as written currently.

man-roxygen/exe_in_path.R
```diff
@@ -0,0 +1,2 @@
+#' @param exe_in_path logical. If TRUE, will look for exe in the PATH. If FALSE,
```
k-doering-NOAA, line 1: I think this can be deleted, because it isn't used as a template anywhere?

vignettes/r4ss-intro-vignette.Rmd
```diff
@@ -94,117 +113,139 @@ If only plots of catch and discards were desired, the user could call:
...
-# Read in ctl file. Note that the data fileR object is needed so that SS_readctl
-# assumes the correct data structure
-ctl <- r4ss::SS_readctl(file = file.path(new_mod_path, start$ctlfil),
-                        verbose = FALSE,
-                        use_datlist = TRUE, datlist = dat)
-fore <- r4ss::SS_readforecast(file = file.path(new_mod_path, "forecast.ss"),
-                              verbose = FALSE)
-# can also read in wtatage.ss for an empirical wt at age model using
-# r4ss::SS_readwtatage()
+inputs <- r4ss::SS_read(dir = new_mod_path)
```
k-doering-NOAA, line 163: Great change!

vignettes/r4ss-intro-vignette.Rmd
```diff
@@ -94,117 +113,139 @@ If only plots of catch and discards were desired, the user could call:
...
-r4ss::SS_writectl(ctl, outfile = file.path(new_mod_path, start$ctlfile),
-                  overwrite = TRUE, verbose = FALSE)
-r4ss::SS_writeforecast(fore, dir = new_mod_path, file = "forecast.ss", 
-                       overwrite = TRUE, verbose = FALSE)
+r4ss::SS_write(inputs, dir = new_mod_path, overwrite = TRUE)
 ```
 
-If you make changes to the input model files that render the file unparsable by Stock Synthesis, the `SS_write*()` functions may throw an error (and hopefully provide an informative message about why). However, it is possible that an invalid Stock Synthesis model file could be written, so the true test is whether or not it is possible to run Stock Synthesis with the modified model files.
+If you make changes to the input model files that render the file
+unparsable by Stock Synthesis, the `SS_write()` functions may throw an
```
k-doering-NOAA, line 224: grammar: functions -> function

R/getADMBHessian.R
kellijohnson-NOAA: I wish that it was a single file path in the instances where there is only one file path of interest.

k-doering-NOAA (commented):
Hi @iantaylor-NOAA , I know this is still a draft, but it looks good (and like you did a ton of work, thank you!) I made a few comments to consider. As for the 2 questions you had:

1. I think using `...` is a good design approach, especially for options that users won't want to modify very often. I like that the documentation notes some things you may want to pass in ...
2. I agree that this is a good idea, and can be done as a second chunk of work to make review easier. Perhaps open an issue as a reminder? It seems ok to leave run() as it is in the PR, though.

R/SS_doRetro.R
iantaylor-NOAA: Good idea. I went ahead with a global find/replace for "1:length(" to "seq_along(" which found 132 instances.

R/SS_output.R
iantaylor-NOAA: good idea

R/SS_tune_comps.R
iantaylor-NOAA: Done, thanks for the reminder.

R/check_exe.R
iantaylor-NOAA: Done.

R/copy_SS_inputs.R
iantaylor-NOAA: Doen.

R/getADMBHessian.R
iantaylor-NOAA: That makes sense. Thanks for bringing this up. I have now deprecated both `dir` (formerly `File`) and `FileName` and merged them into a new single input `hesfile` in commit bd3e7e4708061a0274c8ca53a7bcaad4cc61264a. We could do a generic `file` if you prefer, but it will only ever be one type of file for this function.

I haven't tried to apply this logic elsewhere but can watch out for it as I see it.

R/run.R
iantaylor-NOAA: Good catch. I revised the whole paragraph in addition to making this change.

R/run.R
iantaylor-NOAA: done

R/run.R
iantaylor-NOAA: much improved, thank you.

iantaylor-NOAA: also, copying some of this to the @description for check_exe()

man-roxygen/exe.R
iantaylor-NOAA: Good idea to revise this.

R/SS_RunJitter.R
```diff
@@ -26,7 +20,13 @@
 #' @param init_values_src Either zero or one, specifying if the initial values to
 #'   jitter should be read from the control file or from the par file, respectively.
 #'   The default is `NULL`, which will leave the starter file unchanged.
+#' @template exe
+#' @template verbose
+#' @param ... Additional arguments passed to r4ss::run(), such as
```
kellijohnson-NOAA, line 25: Wrap function name
```
[r4ss::run()]
```
and if R CMD check won't let you use the :: for a function within the package that you are building itself, then `r4ss::run()`.

R/SS_RunJitter.R
```diff
@@ -26,7 +20,13 @@
...
+#' `extras`, and `show_in_console`.
```
kellijohnson-NOAA, line 26: no comma needed for a list of 2.

R/SS_RunJitter.R
kellijohnson-NOAA: This comment carries forward to additional uses of `...`

R/SS_RunJitter.R
iantaylor-NOAA: good catch, fixed by adding forgotten 3rd element to list

R/SS_RunJitter.R
iantaylor-NOAA: Done throughout. 
I was happy to recently discover shift-ctl-H brings up global find-replace in vscode.

iantaylor-NOAA: Follow-up: the links between help pages created by `[r4ss::run()]` worked just fine in the installed package.

iantaylor-NOAA:
@k-doering-NOAA, I think this PR is now ready. I'm not sure how well GitHub will identify the additional changes made since your review of the draft version of this Pull Request, but the focus has been on the issues that you and @kellijohnson-NOAA identified, as well as issue #429 and finally spell check.

R/check_exe.R
```diff
@@ -1,19 +1,44 @@
...
+#' @param dir The directory where `exe` is located (if not in path).
+#' Defaults to `getwd()` but can be an absolute path, a path relative to
+#' the working directory or a path relative to a directory that's in the
+#' PATH. Can also be a vector of directories.
 #' @template verbose
 #' @author Kelli F. Johnson, Ian G. Taylor
-#' @return A list containing the cleaned `exe` name based on the input
-#' argument and the path (or paths if `dir` is a vector) to where it is found.
+#' @return A list containing `$exe` and `$path`.
+#' `$exe` is the cleaned `exe` input file. Windows systems will include
```
k-doering-NOAA, line 15: Reword

```
 `$exe` is the cleaned `exe` input file
```
to
```
 `$exe` is the cleaned version of `exe`
```

k-doering-NOAA: Also clarify if this should be a single value or a vector? I wasn't sure...

R/check_exe.R
```diff
@@ -28,52 +53,62 @@ check_exe <- function(exe = "ss", dir = getwd(), verbose = FALSE) {
...
+        path_to_exe <- ""
       } else {
-        # if not in path or specified directory, create error
-        stop(exename, " not found in ", dir[idir], " nor in the path.")
+        if (verbose) {
+          message("Executable found in PATH at ", path_to_exe)
+        }
+        # if found in PATH and file size is large enough,
+        # then normalize path and remove exe name from the end
+        # e.g. convert "C:\SS\SSB672~1.01_\ss.exe" to "C:/SS/SSv3.30.19.01_Apr15"
```
k-doering-NOAA, line 96: Are the front slashes to back slashes intentional?

R/getADMBHessian.R
```diff
@@ -3,32 +3,40 @@
 #' This function reads in all of the information contained in the
 #' .hes file. Some is needed for relaxing the covariance matrix, while the rest
 #' is recorded and rewritten to file as ADMB expects.
-#' @param dir Directory in which .hes file is located. Defaults to the working
-#'  directory.
-#' @param File Deprecated. Use `dir` instead.
-#' @param FileName Name of .hes file. Defaults to admodel.hes.
+#' @param hesfile Name of .hes file, including the full path (can be
+#' relative to working directory).
```
k-doering-NOAA, line 7: This is great, I find this much more intuitive! Great work.

R/getADMBHessian.R
```diff
@@ -3,32 +3,40 @@
...
   # soft deprecated for now, but fully deprecate in the future.
   if (lifecycle::is_present(File)) {
     lifecycle::deprecate_warn(
       when = "1.46.0",
       what = "getADMBHessian(File)",
-      details = "Please use 'dir' instead"
+      details = "Please use 'hesfile' instead"
     )
-    dir <- File
+    File = getwd()
```
k-doering-NOAA, line 28: Hmm, I think this line `File = getwd()` should be removed, as should line 36. I'm not sure how would be best to construct the output of hesfile... maybe only construct it if both File and FileName are present (using another `if_present() statement)?

R/run.R
```diff
@@ -13,22 +16,25 @@
...
-#' users who want to use the workflow of having an SS3 executable in
-#' their path should name the file something else, such as `ss3` or
-#' `ss_linux`. 
-
-#' @author Ian Taylor, Kathryn Doering
+#' users who want to use the workflow of having SS3 in their PATH should
+#' name the SS3 file something besides `ss`, such as `ss3` or
+#' `ss_linux`.
+#' 
+#' @author Ian G. Taylor, Kathryn I. Doering
```
k-doering-NOAA, line 37: Change `I.` to `L.` (BTW, thanks for standardizing names, I'm pretty bad at remembering to consistently use (or not use) a middle initial.

inst/WORDLIST
```diff
@@ -280,6 +283,7 @@ tagrelease
 tcl
 tgt
 th
+tidyr
```
k-doering-NOAA, line 286: Great to see you adding words to the WORDLIST to make your life easier in the future! This is a practice I need to do :)

R/getADMBHessian.R
iantaylor-NOAA: Good idea. Feel free to check if the change matches what you were thinking.

R/run.R
```diff
@@ -13,22 +16,25 @@
...
+#' @return Returns a table showing which directories had models run and which
 #' had errors like missing executable or Report.sso already present.
 #' The three possible messages are "model run failed", "ran model", and
 #' "unknown run status".
 #'
-#' @description First searches path for the specified executable name
-#' and returns the location of that file if found. If not found in path
-#' it then searches the specified directory. Linux systems have an
+#' @description The `run()` function checks for the executable via
+#' [r4ss::check_exe()]. This involves first checking first checking the
```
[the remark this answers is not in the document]
iantaylor-NOAA: Thank you.

k-doering-NOAA (approved):
Great work, @iantaylor-NOAA ! I support merging this into main, with a few minor wording changes in the documentation. The biggest concern I had was over the how to construct values for deprecated inputs in getADMBHessian, but it should be quick to decide on what to do.

k-doering-NOAA:
Also, there is an option to look only at changes since you last reviewed something on Github, which is a nice feature for big PRs like this one!

R/check_exe.R
iantaylor-NOAA: should have been double back slashes, good catch. I thought I just pasted what I got from "Sys.which('ss')", but somehow lost a slash. Fixed now.

k-doering-NOAA:
All looks, good, thanks @iantaylor-NOAA, I have no other changes to suggest.

Just a note about the failing codecov check - it's failing because the coverage decreased by more than 2% (set in the [codecov.yml file](https://github.com/r4ss/r4ss/blob/c7ecd71b3ecc744c5a133738531eeb8e9ec1b040/codecov.yml#L5). I think this is probably due to taking out some tests for 3.24 code, so I don't think it needs to be worried about.

iantaylor-NOAA:
@k-doering-NOAA, thanks for the info on the codecov check. Those functions were deprecated in a previous pull request #718, but the coverage didn't change, perhaps because the version wasn't yet incremented above the one noted in the calls to lifecycle  https://app.codecov.io/gh/r4ss/r4ss/pull/718. Looking through the files, it indeed seems to be the read/write 3.24 files that had the big drop in coverage. When we delete those files in some future version it will go back up.

Issue #723, opened 2022-07-22
simplify function names that run models


iantaylor-NOAA:
As discussed in PR #722, we can rename the functions that call the executable as follows to avoid a bunch of inconsistent capitalization. Also `run_SS_models()` has already been renamed to `run()` as part of that PR.

  old | new
  -- | --
  SS_RunJitter() | jitter()
  SS_doRetro() | retro()
  SS_profile() | profile()
  SS_tune_comps() | tune()

kellijohnson-NOAA:
Thanks @iantaylor-NOAA for these proposed changes. I am only slightly worried about the use of `tune()` because it is not very clear what it is doing because there are various parts of the model that can be tuned.

chantelwetzel-noaa:
I had the same thought as @kellijohnson-NOAA but I think all of the other proposed function name changes look reasonable.

iantaylor-NOAA:
We can definitely pick something other than `tune()`. How about `tune_comps()`? We don't typically do any other iterative tuning and if we were to add that functionality, it could be put into the same function.

chantelwetzel-noaa:
That would work for me. Plus, its new naming structure would align with all other proposed changes that just cut out the SS_ part.

iantaylor-NOAA:
**Update and questions:**

I started on this effort by first reading https://lifecycle.r-lib.org/articles/communicate.html#rename-a-function
and watching the Hadley Wickham RStudio 2021 keynote about the {lifecycle} package: https://www.rstudio.com/resources/rstudioglobal-2021/maintaining-the-house-the-tidyverse-built/

Then I renamed SS_profile() to profile() in commit 7e2dfe2 in the [rename_functions_issue723](https://github.com/r4ss/r4ss/tree/rename_functions_issue723) branch, via the following changes:
1. copy contents of `R/SS_profile.R` into `R/profile.R`
2. replace the contents of the `SS_profile()` function with a call to `lifecycle::deprecate_stop()`
3. do a global find/replace for `SS_profile` to `profile` to update the tests cross references in documentation
4. update the NEWS.md file to note the changes

I can repeat these for the 3 other functions that are proposed for renaming in this issue but wanted to give folks a chance to provide feedback on two things that could have been done differently

**Question 1**: In terms of git history, does it make sense to create a new file for the new function name or use `git mv R\SS_profile.R R\profile.R` and then create a new file for the old function name? I think GitHub's blame won't work for either approach, so I'm not sure how much it matters.

[...]

PR #736, opened 2022-08-03
simplify run(), deprecate populate_multiple_folders()

### LINKED ISSUE #733: `run()` fails when `dir` is a vector
https://github.com/r4ss/r4ss/blob/37b89fe4046fd271fa76548779e36b22276a0910/R/run.R#L80

If `dir` of `run()` is a vector the function fails because the previous line resets `dir` to be the directory of the loop. Thus, when it is on iteration two of the loop, the function fails because `dir` only has a length of one.

iantaylor-NOAA:
This issue resolves #733 by removing the functionality of having `run()` work on a vector of directories and deprecates the function `populate_multiple_folders()`. I think it's better practice in general to keep these functions smaller and rely on packages like {purrr} to apply them iteratively. We may also want to explore {furrr} for parallel processing with `run()`: https://furrr.futureverse.org/.


R/run.R
```diff
@@ -74,82 +78,75 @@ run <- function(dir = getwd(),
...
-  # return table of results
-  return(data.frame(dir = dir, results = results))
+      }
+      if (isTRUE(console_output > 0)) {
+        results <- "model run failed"
+      } else if (isTRUE(console_output == 0)) {
+        results <- "ran model"
+      } else {
+        results <- "unknown run status"
+      }
```
kellijohnson-NOAA, line 146: I am not positive but I think you could change this complicated chain of if else statements to the following
```
results <- dplyr::case_when(
  console_ouput == 0 ~ "ran model",
  console_output > 0 ~ "model run failed",
  "unknown run status"
)
```

kellijohnson-NOAA (changes requested):
Most comments are related to documentation and condensing the code. The former is more important than the latter imo.

R/populate_multiple_folders.R
```diff
@@ -39,6 +44,15 @@ populate_multiple_folders <- function(outerdir.old = NULL,
                                       exe.file = "ss.exe",
                                       exe.only = FALSE,
                                       verbose = TRUE) {
+  lifecycle::deprecate_warn(
+    when = "1.46.2",
+    what = "populate_multiple_folders()",
+    details = paste(
+      "This function has been deprecated.",
+      "Please use r4ss::copy_SS_inputs() along with {purrr} or other tools."
```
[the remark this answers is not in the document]
iantaylor-NOAA: Thanks for the purrr and lapply examples. I suspect that nobody is using this function and it will be deleted before long, so I'm going to focus the example-writing on the documentation of copy_SS_inputs() where it will have a more lasting impact.

kellijohnson-NOAA:
One more thought I had was should we get rid of the test for the function if the function is still available to users?

R/run.R
iantaylor-NOAA: Good idea. Looks like final line needs to be `TRUE ~ "unknown run status"`

iantaylor-NOAA:
@kellijohnson-NOAA, thanks for the really helpful review. Sorry I left so many legacy elements in `run()` that should have been gone before this PR was created.
This morning I spent too long trying to create a good example of how to replace `populate_multiple_folders()` that was equally useful and easy and finally gave up and reverted the commit to deprecate that function. Instead I've cleaned up the function to work better with `copy_SS_inputs()` which also needed some cleanup.

If you have time for another look at the changes, your input is appreciated.

R/run.R
```diff
@@ -34,122 +30,120 @@
...
-    # confirm that dir exists
-    if (!dir.exists(dir[idir])) {
-      warning("not a directory:", dir[idir])
-      results[idir] <- "not a directory"
+  # confirm that dir exists
+  if (!dir.exists(dir)) {
+    warning("not a directory:", dir)
+    results <- "not a directory"
+  } else {
+    if (skipfinished & "Report.sso" %in% dir(dir)) {
```
kellijohnson-NOAA, line 79: I think I might have mentioned this before but I believe this should be
```
if (file.exists(file.path(dir,"Report.sso")) && skipfinished) {
```
because you only care about skipfinished if the file actually exists. Furthermore, I think that using file.exists rather than %in% is more robust.

R/run.R
```diff
@@ -34,122 +30,120 @@
...
-        } else {
-          results[idir] <- "unknown run status"
-        }
-        setwd(wd_orig) # needed when using relative paths
-      } # end model run
-    } # end code for exe present
-  } # end loop over directories
-  # return table of results
-  return(data.frame(dir = dir, results = results))
+      setwd(wd_orig) # needed when using relative paths
```
kellijohnson-NOAA, line 144: This seems redundant with line 88 and line 88 is more robust.

man-roxygen/show_in_console.R
```diff
@@ -1,2 +1,2 @@
 #' @param show_in_console Show output in the R console? If FALSE,
-#' then it is saved to a file.
+#' then it is saved to a file at the end of the model run.
```
kellijohnson-NOAA, line 2: Do you  need to have this in a template given that it is only used in a single function, i.e., `run()`. Also, there is no default argument called out in the explanation. Line two should be indented by two spaces. FALSE should be wrapped in back ticks. It is not 100% clear what "it" on line two is.

R/populate_multiple_folders.R
```diff
@@ -11,10 +11,12 @@
...
+#' directories or in exe.dir (if provided)? To copy a new executable
+#' from a central location, use `exe.dir` instead.
+#' @param copy_par Copy any .par files found in the individual directories?
+#' @param exe.dir Optional path to executable to copy into all the subfolders.
 #' @param exe.file Filename of executable to copy into all the subfolders.
-#' A value of NULL will skip copying the executable.
-#' @param exe.only Only copy exe files from exe.dir, don't copy input files
 #' @template verbose
 #' @return Returns table of results indicating which directories were
 #' successfully populated with the model input files and/or executables
```
kellijohnson-NOAA, line 22: Full stop needed at end of line 22.

kellijohnson-NOAA (changes requested):
😆 I did the exact same thing in {ss3sim} where I thought I was going to deprecate something and then it ended up being more work than it was worth. I reviewed the new changes and most of them look good to me except the new argument in `population_mutliple_folders()`. I am not sure that it is needed and if you do decide to keep it, I think it needs to be documented a little bit more thoroughly so users know how it interacts with `exe.dir`. Personally, I think that it can be removed.
I am requesting changes largely because I would like to see the examples actually work on people's machines. As is the example for `populate_multiple_folders()` does not work. I provided code that might work but I am unsure if the `outdir.new` does not already exist, if it will work? Perhaps this leads to an additional comment that `outerdir.new` should be documented more. Users should know if this location needs to already exist.

R/populate_multiple_folders.R
```diff
@@ -11,10 +11,12 @@
 #' @param create.dir Create new outer directory if it doesn't exist already?
 #' @template overwrite
 #' @param use_ss_new Use .ss_new files instead of original inputs?
-#' @param exe.dir Path to executable to copy into all the subfolders.
+#' @param copy_exe Copy any executables found in the individual
+#' directories or in exe.dir (if provided)? To copy a new executable
+#' from a central location, use `exe.dir` instead.
```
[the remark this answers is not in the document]
iantaylor-NOAA: This is a great solution. I was not happy with the two arguments but couldn't think of a better solution like the one you propose.

R/populate_multiple_folders.R
```diff
@@ -26,28 +28,39 @@
 #' populate_multiple_folders(
 #'   outerdir.old = "c:/SS/old_models",
 #'   outerdir.new = "c:/SS/new_models",
```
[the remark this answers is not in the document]
iantaylor-NOAA: Good idea. The `creatdir = TRUE` argument allows it to create the new directory created by tempdir()

iantaylor-NOAA:
Thanks for the additional input. These functions are now simpler, better documented, and with better examples.

iantaylor-NOAA:
@kellijohnson-NOAA, I'm going to go ahead and merge this in spite of being on leave for the next week. I think that the changes fix more things than they are likely to break if I made some additional error and would like to get the improved documentation available to people starting to use the revised functions that changed in #723.

In spite of all the work that this took, I think there will be very little impact on users because `populate_multiple_folders()` is not really used by anyone yet, nor is the option to pass a vector of directories to `run()`.

kellijohnson-NOAA:
thanks @iantaylor-NOAA for all of your work on this. If it makes you feel any better, {ss3sim} now uses `r4ss::run()` 😉 as a result of it.

PR #757, opened 2022-11-18
Add .devcontainer to use r4ss in codespaces

k-doering-NOAA:
Adds a .devcontainer based on the FIMS project, but without the C++ needs. Also downloads dependencies of r4ss in the dockerfile.

### DIFF
--- a/.Rbuildignore
+++ b/.Rbuildignore
@@ -17,6 +17,7 @@ docs
 ^inst/extdata/simple_3.30.01/
 ^inst/extdata/simple_3.30.12/
 ^inst/extdata/simple_3.30.13/
+^\.devcontainer$
 # # temporary additions for CRAN submission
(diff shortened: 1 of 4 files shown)


iantaylor-NOAA (approved):
Looks great. I don't know much about Docker or dev containers or codespaces, but this just works. I'm not yet sure how we will use it, but it's easy to imagine that it may make it easier for folks to contribute to r4ss in the future.

We could leave the PR open for a while (until Monday?) in case anyone watching r4ss is interested in codespaces and wants to comment or contribute to these changes.

k-doering-NOAA:
Sounds good! If folks want to try it out, I think all they have to do is go to the `test-codespaces` branch and make a codespace. (60 hrs of codespaces is now free for all GitHub users)

PR #768, opened 2023-01-31
Partition weighting v3

e-gugliotti-NOAA:
Added lines to only pull Length_comp_error_controls if the nrow > 0 for cases where there it is in the definition list but there is no info in the matrix.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -43,6 +43,7 @@ CumReport.sso
 ParmTrace.sso
 ss.cor
 ss.rep
+test.R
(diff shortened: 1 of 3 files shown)


e-gugliotti-NOAA:
@iantaylor-NOAA I have absolutely no idea why these checks aren't passing. Do you have any insight?

iantaylor-NOAA (approved):
Looks good. Thanks @e-gugliotti-NOAA for the fix.

PR #779, opened 2023-02-28
Section508 v2

iantaylor-NOAA:
This PR is related to #411, but does not necessarily make the r4ss HTML view 100% compliance with Section 508 accessibility rules (next step is to work with someone who does the checks to see how it performs). Nevertheless, these changes get things closer also include important improvements in the diagnostic tables, so it's overdue to get merged.


R/SS_html.R
```diff
@@ -351,11 +351,16 @@ SS_html <- function(replist = NULL,
         file = htmlfile, append = TRUE
       )
       for (i in 1:nrow(plotinfo)) {
-        # default alternative text is caption up to any line break <br>
+        # default alternative text is caption up to any line break
```
kellijohnson-NOAA, line 353: @iantaylor-NOAA having the caption repeated as alternative text makes the individual using screen readers have to listen to the caption twice and is thus seen as less ideal than having no alternative text.

iantaylor-NOAA: Thanks @kellijohnson-NOAA. This change makes things slightly less bad than the status-quo by reducing some multi-sentence alt-text to single sentence, but it's easy to change it to "" instead. Is that what you would suggest?

iantaylor-NOAA: Thanks for the feedback on this, @kellijohnson-NOAA. The alt-text is now "" throughout. I will transfer to #411 the suggestion about populating alt-text with reference to the sections in the report file.

Issue #770, opened 2023-02-01
[Bug]: Composition plotting error


matthewsmith-NOAA:
### Describe the bug

SS version 3.30.20
R4ss version 1.47.0 from github

When running SS_plots plotting proceeds as usual until reaching group 13. In group 13, items 21-24 are produced by ssplotComps and output after which the code crashes and returns "Error in if (CompError == 1) { : argument is of length zero". 

R4ss functions were run on a converged model a few days ago and worked as expected (unknown r4SS version). R4ss was updated yesterday and now produces the above error on the same model output. 

[3.Hybrid.zip](https://github.com/r4ss/r4ss/files/10561624/3.Hybrid.zip)

### To Reproduce

Unzip and extract files. Use ss_output to input report file into R without covar (file was too large to attach) or run the model with ss version 3.30.20 to create covar file. Use SS_plots to generate plots. observe error when script reaches plot group 13.

### Expected behavior

I expected the full suite of plots to be produced, exported to the directory on my PC, and the HTML files to be created.

### Which OS are you seeing the problem on?

Windows

### Which versions of r4ss and SS3 are you seeing the problem on?

r4SS 1.47.0, SS3.30.20

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/r4ss/r4ss#code-of-conduct)

iantaylor-NOAA:
Thanks @matthewsmith-NOAA for reporting this. I introduced this bug earlier this week via an inadequately tested changes related to the in-development version of SS3 where separate Dirichlet-multinomial parameters will be available for discards vs retained fish, where that SS3 change required a different table format.

Your model works for me after the change I just made in commit 6ee9723, so please update r4ss (sometimes also requires restarting R), and let me know if it doesn't work for you. If the fix works, go ahead and close this issue.

iantaylor-NOAA:
Update: I accidentally pushed the fix to the branch I was working on, not "main". That branch is now merged via Pull Request #771, so the fix wasn't available via `remotes::install_github("r4ss/r4ss")` 5 minutes ago when I last posted here, but should now be available.

matthewsmith-NOAA:
Thanks Ian! updated package this morning and plots work as expected. Appreciate the quick turn around on the issue.

lisaailloud-NOAA:
[...]

PR #799, opened 2023-04-19
ISSUE #770 Fix: length 0 CompError 

nathanvaughan-NOAA:
Adds a wrapper for length zero Comp Errors when creating figure legends. Bug is due to missmatch in partition value between database and datainfo dataframes. Further investigation into the cause and implications of this mismatch would be helpful.

### DIFF
--- a/R/SSplotComps.R
+++ b/R/SSplotComps.R
@@ -1466,35 +1466,46 @@ SSplotComps <-
                       CompError <- data_info[sub, "error_type"]
                     }
 
-                    # D-M option 1 (linear)
-                    if (CompError == 1) {
-                      Theta <- as.numeric(replist[["Dirichlet_Multinomial_pars"]][["Theta"]][ParmSelect])
-                      # note: in caption below &#920 = Theta
+                    if(length(CompError)!=0){
+                      # D-M option 1 (linear)
+                      if (CompError == 1) {
+                        Theta <- as.numeric(replist[["Dirichlet_Multinomial_pars"]][["Theta"]][ParmSelect])
+                        # note: in caption below &#920 = Theta
+                        caption_extra <-
+                          paste0(
+                            ".<br>'N input' is the input sample size. ",
+                            "'N adj.' is the sample size after adjustment by the ",
+                            "Dirichlet-Multinomial <i>&#920</i> parameter based on the ",
+                            "formula N adj. = 1 / (1+<i>&#920</i>) + N * <i>&#920</i> / (1+<i>&#920</i>). ",
+                            "<br>For this fleet, <i>&#920</i> = ", round(Theta, 3),
+                            " and the sample size multiplier is approximately ",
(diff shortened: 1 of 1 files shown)


iantaylor-NOAA:
I'm going to hold off on merging this until we can investigate the underlying issue further.

However, @lisaailloud-NOAA, you could install Nathan's fix via 
remotes::install_github("nathanvaughan-NOAA/r4ss")
or
pak::pkg_install("nathanvaughan-NOAA/r4ss")

nathanvaughan-NOAA:
@iantaylor-NOAA  I agree on holding off, I already had lisa install my fork for now figuring that was the best alternative. It looks like the issue is that the SS_Output() is not producing all fleet/partition combinations so some fleets only have partition 0 which is throwing the error. If I get a chance to take a look before you I will let you know what I find.

nathanvaughan-NOAA:
I looked into it a little further and the issue is that the Length_comp_error_controls section of the SS report file (below) does not contain all fleet/partition combinations which is what is assumed in the plotting code. Also, I'm not sure why SS is assigning the below partition values when the input comp data appears to be all set as partition 2. This issue may be happening in SS rather than r4ss??   

Fleet partition mintailcomp addtocomp combM+F CompressBins CompError ParmSelect minsamplesize 
1 0 -1 0.001 0 0 0 0 0.001 #_ Com_GN_1
2 0 -1 0.001 0 0 0 0 0.001 #_ Com_HL_2
3 0 -1 0.001 0 0 0 0 0.001 #_ Rec_CB_HB_3
3 1 -1 0.001 0 0 0 0 0.001 #_ Rec_CB_HB_3
3 2 -1 0.001 0 0 0 0 0.001 #_ Rec_CB_HB_3
4 0 -1 0.001 0 0 0 0 0.001 #_ Rec_PRIV_4
4 1 -1 0.001 0 0 0 0 0.001 #_ Rec_PRIV_4
4 2 -1 0.001 0 0 0 0 0.001 #_ Rec_PRIV_4
5 0 -1 0.001 0 0 0 0 0.001 #_ Rec_SH_5
5 1 -1 0.001 0 0 0 0 0.001 #_ Rec_SH_5
5 2 -1 0.001 0 0 0 0 0.001 #_ Rec_SH_5
8 0 -1 0.001 0 0 0 0 0.001 #_ Srv_SEAMAP_8

iantaylor-NOAA:
Aha, this helps. You are right that this is a change in SS3 where the Age_comp_error_controls table is only reporting rows for fleets with age comps rather than all fleets. The format of that table changed recently for models that use partition-specific weights but I hadn't noticed that it was different even if you don't have partition. Shouldn't be too hard to fix in r4ss.

iantaylor-NOAA:
Alternative solution was explored in #801 but closing both PRs because I think the solution in SS3 itself was adequate.

PR #801, opened 2023-04-19
fixes composition plotting error #770

iantaylor-NOAA:
As discussed in #770 there was a bug in the read of the comp data specification table for the format used in 3.30.21. This should fix it.


iantaylor-NOAA:
@e-perl-NOAA thanks for the nudge to resolve these lingering PRs. Looking back at the #770 discussion, it appears that there may be some 3.30.21 models which still have errors with the existing r4ss code, but we've been getting back since April 2023 without needing this PR merged, so I'm just going to close it and trust that anyone with a model that creates an issue can upgrade to a newer SS3 version if they haven't already.

Issue #785, opened 2023-03-03
[Bug]: no shading in plots of unsexed fish


iantaylor-NOAA:
### Describe the bug

Composition data polygons for unsexed fish are empty rather than shaded as shown in the figure below. I have no idea when this originated but it wasn't intended to be this way.

### To Reproduce

Run `SS_plots()` for a model with unsexed fish.

### Expected behavior

Shaded polygons with a slightly different shade of gray for unsexed than males or females.

### Screenshots

![image](https://user-images.githubusercontent.com/4992918/222801656-6aa2482c-f9be-4e69-91e1-b51fb5faf882.png)

### Which versions of r4ss and SS3 are you seeing the problem on?

current "main" branch

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/r4ss/r4ss#code-of-conduct)

iantaylor-NOAA:
This issue is very mysterious.  I looked into it and can't figure out what's going on. The empty polygons only appear when writing to png, not pdf or the default R graphics device. Furthermore, in the figure below I've added some extra polygons using the 3 shades of gray used for the main plot and those turn out fine. I'm going to chalk this up to a weird quirk of base R graphics and trust that we can live with unfilled polygons and maybe this will go away in some future R version. We all have better things to do than worry about this.
![image](https://user-images.githubusercontent.com/4992918/222857507-4e8f6e57-0f62-4c77-930e-23a61e1b6c7b.png)

iantaylor-NOAA:
This issue has been traced back to a relatively harmless issue in SS3 itself (see https://github.com/nmfs-stock-synthesis/stock-synthesis/issues/435) where duplicate comp data were being written to CompReport.sso leading to two polygons on top of each other being drawn (with no space in between for shading). PR #804 adds a step to `SS_output()` which removes any duplicate observations (with a message) as a work around for the 3.30.20 and 3.30.21 which have these observations and should make the plots look normal again.

Thanks to @chantelwetzel-noaa for help finding the source of the problem.

PR #804, opened 2023-04-24
fix: work-around for duplicated unsexed comps

iantaylor-NOAA:
Fixes issue #785.
Thanks to @chantelwetzel-noaa, I figured out that this is an issue with the SS3 output as noted here: https://github.com/nmfs-stock-synthesis/stock-synthesis/issues/435.


chantelwetzel-noaa:
Thank you @iantaylor-NOAA!

PR #812, opened 2023-05-12
Invisible return of logical from `copy_SS_inputs()`

kellijohnson-NOAA:
Small trivial changes to copy_SS_inputs where
* return of logical, TRUE or FALSE, is now invisible so users do not have to assign the function call to a dummy object to suppress the output 
* updated some documentation spacing and added a full stop to the end of an explanation
* updated the documentation for the recursive argument because I didn't understand what it was doing
* added an example that all users can run on their machine

@iantaylor-NOAA I added you as a reviewer to merge this in but you shouldn't actually have to review anything, pretty simple changes I mainly just wanted GitHub actions to run through the checks first before bringing it in to main.

### DIFF
--- a/R/copy_SS_inputs.R
+++ b/R/copy_SS_inputs.R
@@ -18,7 +18,9 @@
 #' @param copy_par Copy any .par files found in dir.old to dir.new?
 #' @param dir.exe Path to executable to copy instead of any in dir.old
 #' @template verbose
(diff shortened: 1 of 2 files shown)


iantaylor-NOAA (approved):
Looks good. Thank you @kellijohnson-NOAA for the improvement in the function and especially in the documentation.
I will wait for the checks to pass before merging because that's just good practice.

PR #816, opened 2023-05-26
DRAFT PR: Transfer workflow

e-gugliotti-NOAA:
Still need to delete the "call" workflow. 
@iantaylor-NOAA, the call workflow has ` with: run-codecov: true` where the `run-codecov` is defined in the actual workflow with the following:
`on:
  workflow_call:
    inputs:
      run-codecov:
        required: false
        type: boolean
        default: false
`
Later in the workflow, it references run-codecov in the Test and the Test coverage portions of the workflow. However, run-codecov = True seems to be what the "call" workflow specifies. Do you want the workflow to run test coverage every time because it seems like that is what it is set up to do now and if so, the Test part could be deleted as well as the inputs.


iantaylor-NOAA:
I think that it makes sense to test the code coverage every with all changes to the code in `R` or `tests` (no need when there are changes to the other files, but I don't know how to set that up). 

The benefit of checking the code coverage each time is that if a change to the code makes a significant difference in code coverage (for good or bad), I think it's better to find that out sooner rather than later so you can either consider revising the tests when you see a decrease, or get the positive feedback right away of having improved the coverage.

iantaylor-NOAA (approved):
looks good so far

e-gugliotti-NOAA:
Finally figured out how to get just the commits from that file with Kelli's
help so will redo with that next week.
*Elizabeth F. Gugliotti (she/her)*
---

*Scientific Software Specialist for Stock Synthesis*

ECS Federal contracted to

NOAA/NMFS/National Stock Assessment Program

Mobile: 843-469-1074

e-gugliotti-NOAA:
Okay then I will delete the input portion as well as:
```
- name: Test
env:
   _R_CHECK_CRAN_INCOMING_REMOTE_: false
   if: inputs.run-codecov != true
   run: devtools::test(stop_on_failure = TRUE)
   shell: Rscript {0}
```
and just have it run codecov always.

e-gugliotti-NOAA:
I've created a new branch that brings over the one workflow file and that history instead of all the workflow files history. Will move conversation here to there.

PR #817, opened 2023-05-26
Transfer workflow with history

e-gugliotti-NOAA:
**From draft PR:**
Elizabeth:
Transfer workflow from workflows repo with history and delete the call workflow.
@iantaylor-NOAA, the call workflow has  with: run-codecov: true where the run-codecov is defined in the actual workflow with the following:
```
on: 
  workflow_call:
    inputs: 
       run-codecov: required: 
       false type: boolean 
       default: false 
```
Later in the workflow, it references run-codecov in the Test and the Test coverage portions of the workflow. However, run-codecov = True seems to be what the "call" workflow specifies. Do you want the workflow to run test coverage every time because it seems like that is what it is set up to do now and if so, the Test part could be deleted as well as the inputs.

Ian:
I think that it makes sense to test the code coverage every with all changes to the code in R or tests (no need when there are changes to the other files, but I don't know how to set that up).

The benefit of checking the code coverage each time is that if a change to the code makes a significant difference in code coverage (for good or bad), I think it's better to find that out sooner rather than later so you can either consider revising the tests when you see a decrease, or get the positive feedback right away of having improved the coverage.

Elizabeth:
Okay then I will delete the input portion as well as:
```
- name: Test
env:
   _R_CHECK_CRAN_INCOMING_REMOTE_: false
   if: inputs.run-codecov != true
   run: devtools::test(stop_on_failure = TRUE)
   shell: Rscript {0}
```
and just have it run codecov always.


iantaylor-NOAA (approved):
Addition and deletion of .yml files looks good.
It looks like `git-filter-repo` was related to tranfering a filtered commit history from the other repository, but does it need to be included in the pull request to work or just run locally on your computer?

kellijohnson-NOAA:
Just a note that `git rm <filename>` removes the file but it is still in the history. This doesn't really matter for such a small file but does matter when files are large and accumulate over time. @e-gugliotti-NOAA I am happy to show you how to delete the file and its history and repush to this branch.

e-gugliotti-NOAA:
@kellijohnson-NOAA That would be great! I have searched online and found this `git filter-branch --index-filter 'git rm -rf --cached --ignore-unmatch path_to_file' HEAD` but I'm not sure if it's correct.

e-gugliotti-NOAA:
Thanks to @kellijohnson-NOAA, I was able to remove the git-filter-remove file properly.

e-gugliotti-NOAA:
Squashed the last 2 commits to remove the separate commit that removed the run on branch portion to test action.

PR #824, opened 2023-06-08
Fix retro to use endyr + 1 for biomass-based quantities

kellijohnson-NOAA:
Thanks to @chantelwetzel-noaa for noticing that the {r4ss} end year of the model is really the end year of the data and that we need to have retrospective calculations related to biomass be end year of the model + 1 because we want to report end of year biomass, where Stock Synthesis calculates beginning of the year biomass. Fishing mortality ($F$) is still based on the end year of the model but all other quantities are now end year + 1. {nwfscDiags} will work just fine as is with these changes. If users actually want to compare the "updated" output with previous calculations they could run the function twice with different inputs to endyrvec.
Commit summary that I believe should be brought in as three separate commits rather than squashed and merged:
1. Reformatted old documentation
2. Styled the file, i.e., wrapped at 80 characters
3. Actually made changes (including adding a test to ensure the length of `endyrvec` is the same as the number of models in `summaryoutput`.

I tested this with the retrospective runs on the server available for the northern copper model.
```
runs <- purrr::map(
  c("../../6.8", paste0("retro-", 1:20)),
  SS_output,
  verbose = FALSE, printstats = FALSE
)
aa <- SSmohnsrho(SSsummarize(runs, verbose = FALSE))
bb <- SSmohnsrho(SSsummarize(runs, verbose = FALSE), endyrvec = 2021:2000)
```
all values in `bb` matched the version of the output currently found in main and the values of F found in main matched the values of F in `aa`.

### DIFF
--- a/R/SSmohnsrho.R
+++ b/R/SSmohnsrho.R
@@ -26,7 +26,7 @@
 #'   is to use the `startyr` of the reference model.
 #' @template verbose
 #'
-#' @author Chantel R. Wetzel and Carey R. McGilliard
+#' @author Chantel R. Wetzel, Carey R. McGilliard, and Kelli F. Johnson
 #' @references
 #' * Hurtado-Ferro et al. 2015. Looking in the rear-view mirror: bias and
 #'   retrospective patterns in integrated, age-structured stock assessment
--- a/R/SSmohnsrho.R
+++ b/R/SSmohnsrho.R
@@ -1,33 +1,35 @@
-#' Calculate Mohn's Rho values for select quantities
+#' Calculate Mohn's rho values for select quantities
 #'
 #' Function calculates:
-#' (1) a rho value for the ending year for each retrospective relative to the reference model
-#' as in Mohn (1999),
-#' (2) a "Wood's Hole Mohn's Rho", which is a rho value averaged across all years for each
-#' retrospective relative to the reference model, and
-#' (3) an "Alaska Fisheries Science Center and Hurtado-Ferro et al. (2015) Mohn's rho,
-#' which is the average rho per retrospective "peel".
+#' 1. a rho value for the ending year for each retrospective relative to the
+#'    reference model as in Mohn (1999);
+#' 1. a ``Wood's Hole Mohn's rho'', which is a rho value averaged across all
+#'    years for each retrospective relative to the reference model; and
+#' 1. an Alaska Fisheries Science Center and Hurtado-Ferro et al. (2015) Mohn's
+#'    rho, which is the average rho per retrospective ``peel''.
 #'
-#'
-#' @param summaryoutput List created by `SSsummarize`. The expected order for the
-#' models are the full reference model, the retro -1, retro -2, and so forth.
-#' @param endyrvec Single year or vector of years representing the
-#' final year of values to show for each model.
+#' @param summaryoutput List created by [SSsummarize()]. The expected order for
+#'   the models are the full reference model, the retro -1, retro -2, and so
+#'   forth.
+#' @param endyrvec Single year or vector of years representing the final year of
+#'   values to show for each model.
 #' @param startyr Single year used to calculate the start of the Wood's Hole
-#' Mohn's Rho value across all years. Defaults to startyr of reference model.
+#'   Mohn's rho value across all years. Defaults to `startyr` of reference
+#'   model.
 #' @template verbose
 #'
-#' @author Chantel R. Wetzel and Carey McGilliard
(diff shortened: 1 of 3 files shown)


R/SSmohnsrho.R
```diff
@@ -1,135 +1,183 @@
...
+  mohn.out[["WoodHole_Rec.all"]] <- sum(mohnRec.all)
+  mohn.out[["WoodHole_Bratio.all"]] <- sum(mohnBratio.all)
+  mohn.out[["WoodHole_F.all"]] <- sum(mohnF.all)
 
-    mohn.out[["WoodHole_SSB.all"]] <- sum(mohnSSB.all)
-    mohn.out[["WoodHole_Rec.all"]] <- sum(mohnRec.all)
-    mohn.out[["WoodHole_Bratio.all"]] <- sum(mohnBratio.all)
-    mohn.out[["WoodHole_F.all"]] <- sum(mohnF.all)
+  # Alaska Fisheries Science Center and Hurtado-Ferro et al. (2015) Mohn's rho
+  # https://www.afsc.noaa.gov/REFM/stocks/Plan_Team/2013/Sept/Retrospectives_2013_final3.pdf
```
kellijohnson-NOAA, line 175: I am noting here that this link is broken and I have already sent an email to try and get an updated link that we can included in the roxygen documentation using `\href{new URL}{Alaska Fishery Science Center}`.

kellijohnson-NOAA:
Spelling error output is
```
DESCRIPTION does not contain 'Language' field. Defaulting to 'en-US'.
Error: Error: Spelling errors found.
  WORD   FOUND IN
Need to ignore some words? Read up on using the WORDLIST file:
https://docs.ropensci.org/spelling/reference/wordlist.html
Execution halted
jmsc   SSmohnsrho.Rd:66
Error: Process completed with exit code 1.
```
which is from a url given for a doi, i.e., `https://doi.org/10.1006/jmsc.1999.0481.` How do we get the spelling check to ignore urls?

R/SSmohnsrho.R
```diff
@@ -1,135 +1,183 @@
...
 #' Function calculates:
-#' (1) a rho value for the ending year for each retrospective relative to the reference model
-#' as in Mohn (1999),
-#' (2) a "Wood's Hole Mohn's Rho", which is a rho value averaged across all years for each
-#' retrospective relative to the reference model, and
-#' (3) an "Alaska Fisheries Science Center and Hurtado-Ferro et al. (2015) Mohn's rho,
-#' which is the average rho per retrospective "peel".
+#' 1. a rho value for the ending year for each retrospective relative to the
+#'    reference model as in Mohn (1999);
+#' 1. a ``Wood's Hole Mohn's rho'', which is a rho value averaged across all
```
iantaylor-NOAA, line 6: It's new to me that you can create a numbered list without actually applying the correct numbering. Super useful.

R/SSmohnsrho.R
iantaylor-NOAA: Thank you. I saw it was broken but couldn't be bothered to find a new link.

iantaylor-NOAA (approved):
Looks good. Thank you @kellijohnson-NOAA.

I don't know how to exclude URLs in general from the spell check, but we don't have too many URLs and it's easy enough to run `spelling::update_wordlist()` which added `jmsc` to WORDLIST and got the test passing again. I also added you as co-author of the function (even if you hadn't made a change to the code, improving the documentation is a vital form of authorship as well).

iantaylor-NOAA:
@kellijohnson-NOAA should we leave this open until you get a response to inquiry about broken URL or squash and merge now and fix it later?

kellijohnson-NOAA:
I can always add the link later no reason to wait.

Though you might want to just rebase and merge rather than squash and merge so the formatting changes in the first two commits do not muddy up future lurkers from being able to hone in on the actual differences in commit 3. If you want I can squash the third commit with your latest two into a single commit and force push to this branch before you merge. All up to you given how you want the commit history to look.

R/SSmohnsrho.R
kellijohnson-NOAA: I am not sure if that functionality works for lists with (1) format, which is not standard in markdown, which is why I switched it to 1.

iantaylor-NOAA:
Thanks again @kellijohnson-NOAA. The commit message in squashed commit (009ba9c667e73508ce2bcaac1b075d74ba8ce7fb) is clear enough and it's not that hard to figure out the substantive change when you know what you're looking for, so I went with squashing everything. It's not worth spending more time on this.

PR #831, opened 2023-06-13
Fix call-r-cmd-check.yml to call the new workflow path

e-gugliotti-NOAA:
The call-r-cmd-check.yml workflow needed to be updated to the new path for the r4ss-extra-tests.yml which is in the r4ss/r4ss repo and not the nmfs-stock-synthesis/workflows repo.


iantaylor-NOAA (commented):
Looks good, thank you

PR #833, opened 2023-06-13
add workflow_call to r4ss-extra-tests.yml

e-gugliotti-NOAA:
fix this workflow to be reusable by adding workflow_call under on:


iantaylor-NOAA (approved):
Looks like it's working. Thank you @e-gugliotti-NOAA.

PR #836, opened 2023-06-24
Add sub-area features to ES tables

chantelwetzel-noaa:
This pull request has a combination of minor fixes and new code additions:

Fix
1. The `SSPlotYield` function is now revised to use `refpoints` function input to allow user control over which lines to include in the yield curve plot.
2. Correct column name from landings to catches if the model does not include a retention curve which implies that the user accounted for discarded fish in the input catches.

Feature
1. Add function input `add_text` to the `SSExecutiveSummary` function that allows users to specify a model area name which will be included in the captions for the executive summary tables.


R/SSexecutivesummary.r
```diff
@@ -260,25 +265,41 @@ SSexecutivesummary <- function(replist,
...
 
       write.csv(es.a, file.path(csv.dir, csv_name), row.names = FALSE)
-      caption <- c(
-        caption,
-        "Recent landings by fleet, total landings summed across fleets, and the total mortality including discards."
-      )
+      if (is.null(add_text)) {
+        caption <- c(
+          caption,
+          "Recent landings by fleet, total landings summed across fleets, and the total mortality including discards."
```
kellijohnson-NOAA, line 274: Couldn't we just change the default add_text to be `"model"` and have a single caption rather than the `if` `else` notation. So for instance, this caption would be "Recent landings by fleet, total landings summed across fleets, and the total mortality including discards included in the model" and users could pass instead `"northern California sub-area model"` and that would be used to replace model? We would basically just change every caption to have a paste at the at end rather than having if else statements for each caption.

chantelwetzel-noaa: Commit [70be856](https://github.com/r4ss/r4ss/pull/836/commits/70be8568975e37fae6ac5246a0618932544827c0) simplifies the captions based on this suggestion

chantelwetzel-noaa:
Commit [70be856](https://github.com/r4ss/r4ss/pull/836/commits/70be8568975e37fae6ac5246a0618932544827c0) address the review comments from @kellijohnson-NOAA. Additional revisions to table column names and captions to include units based on conversations with @iantaylor-NOAA.

iantaylor-NOAA (approved):
Looks good @chantelwetzel-noaa.
I think that fitting these changes in with something I'm working on for spawning output units would just require changing one line, so it's easier to just merge this now and add those changes in a separate PR.

chantelwetzel-noaa:
@iantaylor-NOAA Within this set of commits I added a `so_units` function input so just make sure what you are doing either revises this approach and/or plays nice with this revision.

Issue #449, opened 2020-12-11
wrapper function to call SS_output, then SS_plots?


k-doering-NOAA:
It occurred to me today that I am constantly calling SS_output, then SS_plots, just in the R console, to check results of a model run. I personally prefer to turn off all the verbose output, which makes the call a little more complicated and I often forget which variables to specify in order to turn off all output to the console.

I wonder whether it is worth having even a higher level wrapper function within the r4ss package to create  plots from SS model files? Something simple like:
```r
run_SS_plots <- function(dir) {
  r4ss::SS_plots(r4ss::SS_output(dir = dir, verbose  = FALSE, printstats = F,
                                 hidewarn = TRUE), verbose = FALSE)
}
```
Although maybe there should a `verbose` = T/F option as well? This would save a bit on typing. 

I'm curious if other people would find this useful? I'm also curious about how other people typically use SS_output then SS_plots. For example, do you always call them together, or do you often inspect the returned object from SS_output somehow? Would this sort of function even be useful for your workflow, or is there some way it could be modified to be more useful?

Please feel free to comment on this and provide details about how you typically call these functions!

kellijohnson-NOAA:
What about just adding an argument to SS_plots() that provides a model directory, and if !missing(dir_model) {SS_output()}?

k-doering-NOAA:
I guess that could work, too! Or I wonder if the `replist` is a string (a path to the model dir) instead of a list, `SS_output` could be called? This would avoid adding a new argument, but it may not be intuitive that replist could be either a list object or a string...

This does seem a good solution because it would avoid creating a new function while still reducing typing for calling the plots.

In addition, it would be helpful if verbosity was controlled with just 1 variable. There are currently 3 variables that determine if output is written to the console or not. I'm no sure if the solution is try to deprecate 2 of these switches, make a new variable to be the main switch for verbosity, or if verbose should be changed to be a main switch?

kellijohnson-NOAA:
when you read in SS_output()$inputs$dir is the directory location, so you could just make a check in SS_plots() to see if the rest of the list is empty and if it is, then read it in using 
```
# power users could call this
SS_plots(replist = list(inputs = list(dir = getwd()))
# this line would go in SS_plots after an if statement
replist <- SS_output(replist$inputs$dir)
```
this could just be a behind the scenes thing that would only minorly need to be documented.

k-doering-NOAA:
I think I would prefer to call something more like
```r
[...]

PR #839, opened 2023-07-06
Get ss exe function added

e-gugliotti-NOAA:
Added get_ss3_exe() function to grab the latest release of the ss3 exe according to the OS if version = NULL, or the user can put in the version tag and pull that version of the exe. Updated the r4ss intro vignette accordingly.
Partial solution to issue [#449](https://github.com/nmfs-stock-synthesis/stock-synthesis/issues/449) in the stock-synthesis repo.


.RData
iantaylor-NOAA, line 1: I think this .RData file was probably added by accident.

R/get_ss3_exe.r
```diff
@@ -0,0 +1,85 @@
+#' Download the Stock Synthesis (SS3) executable
+#'
+#' Downloads the SS3 executable according to specified version and the user 
+#' operating system.
+#'
+#' @param dir The directory that you would like the executable downloaded to.
+#' @param version A character string of the executable version tag to download 
+#' (e.g.'v3.30.20' or 'v3.30.18').
```
iantaylor-NOAA, line 8: Maybe expand `version` to add a sentence like. "A list of tags is available at https://github.com/nmfs-stock-synthesis/stock-synthesis/tags."

R/get_ss3_exe.r
```diff
@@ -0,0 +1,85 @@
...
+    if(!version %in% tags){
+      warning("The version tag that you entered is invalid or not in the right format, 
+              please go to https://github.com/nmfs-stock-synthesis/stock-synthesis/tags 
+              to get a correct version tag or version tag format")
+    } else{
+      tag <- version
+    }
+  }
+  
+  if(is.null(dir)){
```
iantaylor-NOAA, line 45: The function crashes if you try to download to a directory which doesn't exist. I would either add something like
```
  if (!dir.exists(dir)) {
    stop("Directory doesn't exist: ", dir)
  }
```
Or better yet, create the directory. There are various, surely inconsistent approaches to that in other r4ss functions, including this one in `copy_SS_inputs()` where you get to choose whether to allow it to create a new directory, and if so, whether to recursively created nested directories beyond just a single addition:
https://github.com/r4ss/r4ss/blob/main/R/copy_SS_inputs.R#L63-L70.

R/get_ss3_exe.r
```diff
@@ -0,0 +1,85 @@
...
+      warning(
+        "Stock Synthesis binary is not available for 32-bit ",
+        .Platform$OS.type, ".")
+    } else {
+      url <- paste0(
+        "https://github.com/nmfs-stock-synthesis/stock-synthesis/releases/download/",
+        tag, "/ss_win.exe")
+      utils::download.file(url, destfile=file.path(dir, "ss3.exe"))
+      download_location <- file.path(dir, "ss3.exe")
+      return(paste0("The stock synthesis executable ", tag, " was downloaded to: ",
```
iantaylor-NOAA, line 62: Here and elsewhere I think it would be better to change this `return()` to `message()` and then follow it with `return(TRUE)`. That way if this function were being called within some more complex code it would be easier to confirm that it got to the end.

To be extra fancy, maybe it would be useful to expand the first part of the message from "The stock synthesis executable..." to "The stock synthesis executable **for Windows**...", and likewise "for Mac" and "for Linux". That way folks will know that the function was designed for all three operating systems and not have to wonder if it's another case of a Windows-centric world making their non-Windows life more difficult.

inst/WORDLIST
```diff
@@ -141,7 +141,6 @@ endyr
 eol
 et
 exe
-faq
```
iantaylor-NOAA, line 144: thanks for running `spelling::update_wordlist()`!

tests/testthat/test-get_ss3_exe.R
```diff
@@ -0,0 +1,26 @@
+# example_path <- system.file("extdata", package = "r4ss")
```
iantaylor-NOAA, line 1: Thanks for creating a test!

vignettes/r4ss-intro-vignette.Rmd
```diff
@@ -234,6 +234,23 @@ If you need help troubleshooting `SS_read()` or `SS_write()` and the
 associated functions for each model file, or would like to report a bug,
 please [post an issue in the r4ss repository](https://github.com/r4ss/r4ss/issues).
 
+### Download the Stock Synthesis executable from GitHub
```
iantaylor-NOAA, line 237: Thanks for adding to the vignette!

iantaylor-NOAA (approved):
@e-gugliotti-NOAA, this is an ideal pull request in that it provides a really useful addition to the package, that new function is well documented, described in the vignette and you've added a test is added and even the updated the spelling list.

I made a few minor suggestions because it's hard to not want to make even a really good thing a tiny bit better.

R/get_ss3_exe.r
e-gugliotti-NOAA: going with the stop if directory doesn't exist. I'd rather people be downloading to a known place so it's more easily findable.

Issue #842, opened 2023-07-08
[Bug]: file permission to execute "ss3" downloaded by get_ss3_exe on non windows OS


yukio-takeuchi:
### Describe the bug

I am not sure if it is really a bug. I'd rather to make a clarification of the "expected" behavior of get_ss3_exe function.

For  linux and macOS, ss3 downloadable from https://github.com/nmfs-stock-synthesis/stock-synthesis doesn't have file permission to execute while in the case of windows ".exe" extension  immediately allow to "ss3" to be executable.

### To Reproduce

On macOS or linux

```
➜ Rscript -e "r4ss::get_ss3_exe()"
No directory provided, the executable will be downloaded to the
            working directory
trying URL 'https://github.com/nmfs-stock-synthesis/stock-synthesis/releases/download/v3.30.21/ss_osx'
Content type 'application/octet-stream' length 4139312 bytes (3.9 MB)
==================================================
downloaded 3.9 MB

The stock synthesis executable for Mac v3.30.21 was downloaded to: /Users/yukiot/test_get_ss3_exe/ss3

~/test_get_ss3_exe took 3s
➜ ./ss3
zsh: permission denied: ./ss3
❯ ls -lh ss3
-rw-r--r--  1 yukiot  staff   3.9M Jul  8 15:55 ss3
```

### Expected behavior

If it is better to make "ss3" file executable immediately after download on linux and macOS, it may be better to add an additional code to change file permission in the case of linux or macOS.

If so 

a line 

```
Sys.chmod(paths = file.path(dir, "ss3"), mode =  "0700")
```

may work to make "ss3" executable.

But alternative option may be to include an addtional argument of get_ss3_exe to specify the desirable file permission of "ss3", e.g. "mode = '0777'". Then a new line like

```
Sys.chmod(paths = file.path(dir, "ss3"), mode =  mode)
```

### Which OS are you seeing the problem on?

Mac, Linux

### Which versions of r4ss and SS3 are you seeing the problem on?

latest r4ss@main

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/r4ss/r4ss#code-of-conduct)

e-gugliotti-NOAA:
That is a great point @yukio-takeuchi. I will get working on making the exe immediately usable to linux and max users today.

iantaylor-NOAA:
Fixed by @e-gugliotti-NOAA via #844.

PR #844, opened 2023-07-11
Get ss3 exe linux/ios fix

e-gugliotti-NOAA:
- Resolves issue #842.
- Created a test to run the simple_small model with the downloaded exe to make sure that it works and the new Report.sso file has file.info() where the date is the date that the test is run.
- Default permissions is 700 for mac and linux.
- Needed to use a different download method for windows exe because the method used for the mac and linux exe wasn't working. Also changed the name of windows exe to ss_win.exe instead of ss3.exe because that seemed to be messing things up too.


yukio-takeuchi:
Thank you for quick change to modify file permission for linux and macOS.
On windows I think the cause that ``utils::download.file`` fails to download an "executable"  executable may be default argument of mode = "w" of download.file. This should be mode="wb" to tell the file to download is binary. Otherwise download.file function assume the file is text file so that downloaded file will be broken.

iantaylor-NOAA (approved):
Thanks @e-gugliotti-NOAA.

I had issues with installing the `downloader` package yesterday, but now that we're back to the `utils::download.file()` thanks to the input from @yukio-takeuchi, everything is working fine for me. I tested on Windows and Mac computers and no file permission issues on the Mac.

I note that the use of "ss3.exe" or "ss3" for the executable name differs from the default `ss` in `r4ss::run()` but I think it makes sense to make this change to align with other uses of "SS3" instead of "SS" throughout the documentation.

PR #846, opened 2023-07-12
Change default exe name from "ss" to "ss3"

iantaylor-NOAA:
Changes default executable name from "ss" to "ss3" throughout the {r4ss} package. This was inspired by the addition of the new `r4ss::get_ss3_exe()` function thanks to @e-gugliotti-NOAA, but I think was an overdue shift to align with use of "SS3" or "ss3" in many other places.

I think when we switched to compiling the SS3 executables via github actions, it was necessary to switch from "ss" to "ss_win", "ss_osx" and "ss_linux", but the addition of the new `r4ss::get_ss3_exe()` function (via #839 and #844) allows users to easily install the executable with name "ss3" on all these operating systems without having to rename the file.

I created this branch as a branch from [get_ss3_exe-linux/ios-fix](https://github.com/r4ss/r4ss/tree/get_ss3_exe-linux/ios-fix) rather than main, but I think we should squash and merge #844 first and then this one so they remain separate in the commit history rather than merge this into that branch as a first step and then squash and merge.


iantaylor-NOAA:
The code coverage test shows a significant drop in coverage, indicating that the tests are not all running as expected after this change. I can debug at the start of August after the petrale sole review next week but probably not before then.

e-gugliotti-NOAA:
@iantaylor-NOAA I can try to help with this when I get out from underneath all the work I need to catch up on after I get back from the ESIP conference.

e-gugliotti-NOAA:
@iantaylor-NOAA I don't know why this started failing now, but it also failing locally. I'm not super knowledgeable about the jitter fraction.

iantaylor-NOAA:
@e-gugliotti-NOAA, the `r4ss::jitter()` function is tries to change a starter file line from 0 to 0.1 but is supposed to not run if it can't find the executable in this section: https://github.com/r4ss/r4ss/blob/ae2b0a5982d761cbbd96582c5266b7123445da53/tests/testthat/test-runs.R#L136-L142 such that the starter file value would remain at 0. In this case it seems that the `skipexe` variable calculated above in https://github.com/r4ss/r4ss/blob/ae2b0a5982d761cbbd96582c5266b7123445da53/tests/testthat/test-runs.R#L121-L122 returned TRUE, but the jitter function was able to run in spite of that. I'm not sure what would have caused that to go wrong now but maybe this additional information will be enough to figure out how to look into it deeper if you have time.

e-gugliotti-NOAA:
@iantaylor-NOAA I believe that I needed to add to the test-get_ss3_exe to delete the executable after every test is run. I also rebased since there were some changes made to main that affect this.

k-doering-NOAA:
my (not well checked) guess of what might be going on is that code cov was running with tests that required exes, but something about changing the name means those tests are not run, so that the code cov appears lower. Happy to help see if this is the case!

iantaylor-NOAA:
@e-gugliotti-NOAA thanks for circling back about this pull request. I hadn't realized that one of your previous commits had resolved the code coverage issue, but all checks are passing and coverage remains similar to before.

It's good to have the package now be internally consistent and work well with the new function `r4ss::get_ss3_exe()` and the now-common use of SS3 throughout the documentation for the modeling platform. Obviously some instances of "ss" remain, including in the name "r4ss", but I think the stand-alone occurrences are all now "ss3".

PR #854, opened 2023-07-21
Use return instead of message for get_ss3_exe

### LINKED ISSUE #852: `get_ss3_exe()` does not return the path
The documentation states that the path of the downloaded exectuable is returned from the function but it is not. Instead `NULL` is returned.
https://github.com/r4ss/r4ss/blob/92821e749d7e72747bd010f6187b500204160128/R/get_ss3_exe.r#L10

After the terminal if else statement the path needs to be wrapped in `return()`, i.e.,
```
return(download_location)
```

e-gugliotti-NOAA:
Go back to return() instead of message() for where exe is downloaded.

@iantaylor-NOAA I ran devtools::document() and the man/SSplotSPR.Rd changed, I guess you did updates to the R file recently? Can you double check that this is right?

Resolves #852

### DIFF
--- a/R/get_ss3_exe.r
+++ b/R/get_ss3_exe.r
@@ -66,7 +66,7 @@ get_ss3_exe <- function(dir = NULL, version = NULL) {
       )
       utils::download.file(url, destfile = file.path(dir, "ss3.exe"), mode = "wb")
       download_location <- file.path(dir, "ss3.exe")
-      message(paste0(
+      return(paste0(
         "The stock synthesis executable for Windows ", tag, " was downloaded to: ",
         download_location
       ))
@@ -78,7 +78,7 @@ get_ss3_exe <- function(dir = NULL, version = NULL) {
       Sys.chmod(paths = file.path(dir, "ss3"), mode = "0700")
       download_location <- file.path(dir, "ss3")
 
-      message(paste0(
+      return(paste0(
         "The stock synthesis executable for Mac ", tag, " was downloaded to: ",
         download_location
       ))
@@ -88,7 +88,7 @@ get_ss3_exe <- function(dir = NULL, version = NULL) {
         utils::download.file(url, destfile = file.path(dir, "ss3"), mode = "wb")
         Sys.chmod(paths = file.path(dir, "ss3"), mode = "0700")
         download_location <- file.path(dir, "ss3")
-        message(paste0(
+        return(paste0(
           "The stock synthesis executable for Linux ", tag, " was downloaded to: ",
           download_location
         ))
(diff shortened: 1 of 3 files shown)


iantaylor-NOAA:
@e-gugliotti-NOAA, thanks for working on this. I forgot to run `devtools::document()` locally after changing `SSplotSPR()` and there's an issue with the style code and document github action at the "Install curl dependency" step: https://github.com/r4ss/r4ss/actions/workflows/call-doc-and-style-r.yml so the Rmd files are not getting updated automatically. I'm hoping that issue will go away.

As for the change to `get_ss3_exe()` I have a minor suggestion that I will add in a review shortly.

iantaylor-NOAA (changes requested):
@kellijohnson-NOAA should chime in on this, but my preference would be to provide separate information as a message and as a return.

That is, I like the status-quo information provided via `message()` because it's in English and readable by humans. I'm guessing that @kellijohnson-NOAA is using this function within some R script where the code doesn't care about grammar or what operating system it is (because it already knows) and just wants the string `download_location`.

Thus, I think it would makes sense to use both functions something like below (where `invisible()` avoids duplicating the message if the user isn't saving the results to some object).
```
message(paste0(
        "The stock synthesis executable for Windows ", tag, " was downloaded to: ",
        download_location
))
return(invisible(download_location))
```

e-gugliotti-NOAA:
@iantaylor-NOAA Happy to make that change!

iantaylor-NOAA (approved):
Looks good to me, thank you @e-gugliotti-NOAA

kellijohnson-NOAA:
Thanks for implementing the return statement so I can use the output as input for another function. I would personally just write one return statement at the end rather than have multiple return statements inside the loops. As written now, if it goes inside the else statement on line 99 then it will end up returning a NULL value. I am not sure if this is intended? Do we want to return the file path that the executable was going to be downloaded to or a NULL because nothing was actually downloaded? I am honestly not sure.

e-gugliotti-NOAA:
I think NULL because I could see it being confusing if it returned the path that it should be downloaded but also provides a warning. As someone using the return statement in another function, @kellijohnson-NOAA would there ever be a use case for you to have the path returned even if the exe wasn't downloaded?

iantaylor-NOAA:
@kellijohnson-NOAA thanks for the input.
How about we change `warning()` to `stop()` on line 99 and then it's more obvious that the function failed and won't return anything?

@e-gugliotti-NOAA, I think @kellijohnson-NOAA is right--could you replace the three identical `return()` statements with just one at the end? Sorry I didn't think to suggest that before.

e-gugliotti-NOAA:
@kellijohnson-NOAA and @iantaylor-NOAA does the new commit look good to you?

iantaylor-NOAA:
Looks good to me. I tested it out and thought of two minor tweaks which I added in 02c9e35 (removing some extra spaces in a message and changing "full file path" to "file path" in the `@return` description because if you provide a path relative to the base model, you get that same thing back (which I think is appropriate).

We probably should wait for the latest checks to complete but we can merge at time after that.

PR #865, opened 2023-09-26
fix error when comparing EWAA models with no comps

iantaylor-NOAA:
There was an error related to missing `SpawnOutputUnits` when comparing models which used empirical weight-at-age but had no comp data. This should allow the comparisons to go forward.


iantaylor-NOAA:
Minor change is working.

PR #868, opened 2023-10-02
change jitter for loop to future_map for parallelization, first try

okenk:


R/jitter.R
```diff
@@ -190,3 +203,77 @@ jitter <- function(dir = getwd(),
...
+    # delete jitter model directory
+    unlink(jitter.dir, recursive = TRUE)
+    return(like)
+  } else {
+    unlink(jitter.dir, recursive = TRUE)
+    if (verbose) warning("No Report.sso file found from run ", i)
+  }
+}
+
+plan(sequential)
```
iantaylor-NOAA, line 271: could you move these lines to the `@examples` section at the top? Currently it tries to run these when loading the package.

iantaylor-NOAA (changes requested):
@okenk, this is excellent, thank you.
There's lots of overhead to file management in SS3, so examples that take a short time to run won't see too much benefit. The production assessment models that are most important to parallelize are often at least 10 minutes and will see a big benefit from parallelization. Moving to cloud computing where things can be scaled up further will presumably further increase that benefit.

We could either require availability of {furrr} or make it optional.

To require, run `usethis::use_package("furrr")`
to keep it optional, `usethis::use_package("furrr", type = "Suggests")`.
Either command will edit the DESCRIPTION file and running `devtools::document()` will add it to the dependency to NAMESPACE.

PR #869, opened 2023-10-03
update vignette with SSplotComparison plots

e-gugliotti-NOAA:
For some reason the SSplotComparisons isn't showing the 2nd subplot when I Knit the vignette so I had to grab the .png file to show the plot. If I do subplot = 2:3 it will show the 3rd one but not the second. However, when I knit the chunk in RStudio it shows the plots when running SSplotComparisons. It's so weird.

Please try this locally for you to see if it does the same thing when you Knit the vignette.

### DIFF
--- a/.github/workflows/r4ss-extra-tests.yml
+++ b/.github/workflows/r4ss-extra-tests.yml
@@ -64,16 +64,16 @@ jobs:
       - name: Get the latest SS3 executable and move to expected location
         run: |
           curl https://api.github.com/repos/nmfs-stock-synthesis/stock-synthesis/releases/latest | grep "browser_download_url" | grep -Eo 'https://[^\"]*' | grep "ss_linux" | xargs wget
-          mv ss_linux ss
-          sudo chmod a+x ss
-          cp ss inst/extdata/simple_small/ss
-          rm ss
+          mv ss_linux ss3
+          sudo chmod a+x ss3
+          cp ss3 inst/extdata/simple_small/ss3
+          rm ss3
 
       - name: Get a 3.24 SS executable
         run: |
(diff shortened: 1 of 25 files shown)


iantaylor-NOAA (approved):
@e-gugliotti-NOAA, this looks good. 
Running `devtools::build_vignettes()` locally initially had an issue with 
```
Quitting from lines 243-248 [unnamed-chunk-14] (r4ss-intro-vignette.Rmd)
Error: processing vignette 'r4ss-intro-vignette.Rmd' failed with diagnostics:
cannot open URL 'https://github.com/nmfs-stock-synthesis/stock-synthesis/releases/download/v3.30.21/ss_win.exe'
--- failed re-building 'r4ss-intro-vignette.Rmd'
```
so I temporarily added `eval = FALSE` to the first uses of `get_ss3_exe()`. However, those lines have been there for months without issue, and the later use of that function worked when running again on my computer, so maybe it was just a fluke and I didn't push the changes.

I modified the call to `SSplotComparisons()` slightly to avoid printing the extra plots when you knit in RStudio. Since `knitr::include_graphics()` works, I'm guessing there's no need to plot to the graphics device unless that would make the vignette cleaner, but feel free to modify as you wish.

e-gugliotti-NOAA:
okay so when plot = TRUE, the SSplotComparisons was outputting double plots for you locally when you knit the vignette? I mean that's what it's supposed to do, but it wasn't doing it for me when I knit the vignette for some reason which is why I added the include_graphics bit.

e-gugliotti-NOAA:
Also, I think eval=FALSE should be there for the earlier get_ss3_exe instances since they use new_mod_path and that doesn't exist in that portion of the code

e-gugliotti-NOAA:
@iantaylor-NOAA all things seem to be running well now so unless you have any final thoughts/edits, I'll go ahead and merge

iantaylor-NOAA (commented):
Looks good, go ahead and merge. I think we should do more squash-and-merge which I forgot on the last PR for the change exe name.

PR #870, opened 2023-10-03
Running diagnostics in parallel

okenk:
This updates `retro()` and `jitter()` to use `furrr` functions instead of for loops. I held off on `profile()` because there is some serial dependence if `usepar = TRUE`. It seems like a little bit of a headache? For `jitter()`, I moved the contents of the for loop to a new function, `iterate_jitter()` that is NOT exported. For `retro()`, because no changes in file management were required, I got lazy and wrote a VERY long anonymous function.

One thing I wasn't sure about: I used the new anonymous function syntax (e.g., `future_map(1:5, \(x) x^2)`). It looks like {r4ss} only requires R version 3.5.0 though. Since the syntax is newer, I am not sure if it will work? (Can all of the dependencies really be installed on 3.5.0?)

I'm sure there are other things I did poorly, too.

This resolves some of #775.


R/retro.R
```diff
@@ -72,6 +74,13 @@ SS_doRetro <-
 #'   endyrvec = endyrvec,
 #'   legendlabels = paste("Data", 0:-5, "years")
 #' )
+#' 
+#' ## run retrospectives in parallel
+#' future::plan(future::multisession)
```
e-gugliotti-NOAA, line 79: Just a thought:
In my workflow for something else that I am using furrr for, I have:
```
ncores <- parallel::detectCores() - 1 
future::plan(multisession, workers = ncores)
```
I saw it on some tutorial for working with furrr awhile back, but I'm not sure if it's a best practice. The workflow I have is using a 64-core azure VM, but I guess if others are using something as large, it might be worth adding a similar line to the example

R/retro.R
```diff
@@ -97,20 +106,20 @@ retro <- function(dir = getwd(), masterdir = lifecycle::deprecated(),
...
   subdirnames <- paste0(subdirstart, years)
-
+  
   # check for executable
   check_exe(exe = exe, dir = olddir, verbose = verbose)
-
+  
   # loop over retrospective years
-  for (iyr in seq_along(years)) {
+  furrr::future_walk(seq_along(years), function(iyr) {
```
e-gugliotti-NOAA, line 119: I believe that you could just use:
```
furrr:::future_walk(years, function(iyr){}
```
instead of having seq_along(years)
You would also have to change
```
starter[["retro_yr"]] <- years[iyr]
```
to 
```
starter[["retro_yr"]] <- iyr
```
on line 135

e-gugliotti-NOAA: Would also have to change line 120 to be something like:
```
newdir_iyr <- file.path(newdir, subdirnames[which(years==iyr)])
```
if not using seq_along(years)

okenk:
> @okenk, you may want to rebase your branch from the r4ss main branch because the exe name in the tests and in the github action that downloads the exe has been renamed from ss to ss3 (@iantaylor-NOAA just pushed these changes yesterday right before you created this PR). This may be causing the issue in the run-tests workflow.

Thank you! I was very confused why it was failing. I will look into your other recommendations, too.

R/jitter.R
```diff
@@ -61,6 +79,13 @@ SS_RunJitter <-
 #'   jitter_fraction = 0.1, init_value_src = 1
 #' )
 #'
+#' #### Run same jitter in parallel
```
e-gugliotti-NOAA, line 82: Same comment as with the retro function about potentially adding:
```
ncores <- parallel::detectCores() - 1 
future::plan(multisession, workers = ncores)
```

R/jitter.R
```diff
@@ -134,50 +159,23 @@ jitter <- function(dir = getwd(),
...
-      }
-      # rename output files
-      file_increment(i = i)
-    } else {
-      if (verbose) warning("No Report.sso file found from run ", i)
-    }
-  }
-  # Move original files back
-  pattern0 <- dir(pattern = "[a-z_]0\\.sso")
+  likesaved <- furrr::future_map_dbl(Njitter, function(x)
```
e-gugliotti-NOAA, line 171: I believe that you can use the following format instead of function(x)
```
furrr::future_map_dbl(Njitter, ~iterate_jitter(
       i = .x,
       dir = dir,
       printlikes = printlikes,
       exe = exe,
       verbose = verbose,
       ...))
```

e-gugliotti-NOAA: But i've only started using furrr recently myself so please don't take my word for everything here!

R/retro.R
iantaylor-NOAA: @e-gugliotti-NOAA, I'm trying to understand the default number or workers associated with `future::plan(future::multisession)`. It looks like maybe it's all the available cores, so providing an example where the user has control as you suggest makes good sense.
I see here: https://parallelly.futureverse.org/reference/availableCores.html#avoid-ending-up-with-zero-cores-1 a suggestion to use `ncores <- parallelly::availableCores(omit = 1)`.

I was a little concerned about having r4ss depend on a bunch of additional packages, but it looks like {furrr} brings in {globals}, {parallelly}, and {future} which are all part of the well-supported futureverse collection and not that large.

R/retro.R
okenk: Thanks! Honestly, I did not actually write this code. For `retro()` I basically just switched out the `for` loop for `future_walk()`. It seems like it is kind of a six to one, half a dozen to the other situation?

R/retro.R
okenk: I am definitely not a {furrr} expert either! Just enthusiastic about getting this stuff running faster, and potentially moving some of these really slow things like jittering to the cloud. I will add this line to the examples for both functions.

R/retro.R
iantaylor-NOAA: It's probably clunkier code, but I like `iyr` as an index and `years` as a vector so will just leave as-is for now.

iantaylor-NOAA (approved):
Looks good.

PR #877, opened 2023-10-11
fix vignette comparison plot image

e-gugliotti-NOAA:
If you look at the vignette, the image doesn't show up. I created the image and copied it over to the vignette folder so it could be directly referenced from the vignette folder.


iantaylor-NOAA (approved):
Thanks for fixing it @e-gugliotti-NOAA. Looks good.

Issue #878, opened 2023-10-11
[Feature]: add read of new "maximum_ASEL2" output added for next SS3 version


iantaylor-NOAA:
### Describe the feature request.

New table being produced by these lines should be read by `SS_output()`
https://github.com/nmfs-stock-synthesis/stock-synthesis/blob/main/SS_write_report.tpl#L2935-L2937

This information could be helpful to quickly flag any fleets where the interaction of growth and selectivity leads to fish never being fully selected at any age.

Tasks include:
- [ ] Reading the new table in `SS_output()`
- [ ] [optional] Adding any info from the new table to the plot of ASEL2, it's caption, or some other output location.

### Describe alternatives you have considered

NA

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/r4ss/r4ss#code-of-conduct)

iantaylor-NOAA:
@Rick-Methot-NOAA, what was the motivation for adding the new `maximum_ASEL2` table?
The remaining (optional) item on this issue is providing feedback to the user about the results of that table via the caption of the plot showing the ASEL2 values or elsewhere, so any guidance you have on how to present that would be useful.

PR #880, opened 2023-10-16
feat: add read of new "maximum_ASEL2" output #878

iantaylor-NOAA:
- resolves issue #878
- relates to nmfs-stock-synthesis/stock-synthesis#500


e-gugliotti-NOAA:
Tested with recent ss3 exe build and the new table shows up in the output.

e-gugliotti-NOAA (approved):

PR #881, opened 2023-10-16
try new permissions for get_ss3_exe.r

e-gugliotti-NOAA:


e-gugliotti-NOAA:
Fix permissions issue that @okenk ran into when running on an Azure linux vm.

iantaylor-NOAA (changes requested):
I don't know much about the `Sys.chmod()` function in the line above added via PR https://github.com/r4ss/r4ss/pull/844 (in response to suggestions from @yukio-takeuchi in https://github.com/r4ss/r4ss/issues/842), or how that compares to calling "chmod" within "system". However, I'm guessing that we can combine the two things by changing the line above (line 89) from
`Sys.chmod(paths = file.path(dir, "ss3"), mode = "0700")`
to 
`Sys.chmod(paths = dir, mode = "0700")` 
which should modify permissions for all the files in that directory, not just the exe.
Or maybe the problem is that the Azure vm needs to permissions to be not just for the user by the whole system, in which case we could change to 
`Sys.chmod(paths = dir, mode = "0777")`
(see https://en.wikipedia.org/wiki/File-system_permissions#Numeric_notation for more info on the codes).

@okenk, do any of these make sense for the issue you're seeing?

If there's any concern about modifying the permissions for the system not just the user, we could follow one of the earlier options suggested by @yukio-takeuchi and add something like `chmod_mode = "0777"` as an additional argument to the function so folks could change that if needed, but I wouldn't bother unless somebody asks for it.

iantaylor-NOAA:
If applying Sys.chmod() to the directory doesn't also change permissions for the executable within it (I think it should), then adding `recursive = TRUE` should do so.

e-gugliotti-NOAA:
I'll try it out on the codespace I have for this branch.

okenk:
So I figured out JUST enough to get stock synthesis running on a Linux VM, and changing the permissions did help. I could not run `get_ss3_exe()` from my model directory. I got this warning message that I showed to @e-gugliotti-NOAA:
```
The stock synthesis executable for Linux v3.30.21 was downloaded to: /home/kiva-oken/canary-model/ss3
Warning messages:
1: In utils::download.file(url, destfile = file.path(dir, "ss3"), mode = "wb") :
  URL https://github.com/nmfs-stock-synthesis/stock-synthesis/releases/download/v3.30.21/ss_linux: cannot open destfile '/home/kiva-oken/canary-model/ss3', reason 'Permission denied'
2: In utils::download.file(url, destfile = file.path(dir, "ss3"), mode = "wb") :
  download had nonzero exit status
```
I also got an error when running stock synthesis (which I moved myself without {r4ss}), after it read the control file:
```
Error opening temporary gradient file cmpdiff.tmp
```

At this point the permissions for the directory were drwxr-xr-x and for ss_linux were -rwxr-xr-x.

Then, I updated the permissions *for the directory* to be drwxrwxrwx, and stock synthesis ran, so I went on my merry way. 

Oddly, I just created a new folder, which by default had permissions drwxr-xr-x and *did* I successfully run `get_ss3_exe()` (And I did not update {r4ss} to include this pull request.) 

I *think* some weird things happened when I copied a directory over from my laptop using Azcopy.

I can try to look into this more to figure out what the correct fix is.

e-gugliotti-NOAA:
I commented out what I had, and put in something else. These all are working in codespaces, but idk what will work best on the vm.

okenk:
Actually, I think I know what the issue is. I'm not sure what the third column means, but the directory I copied from Azcopy says "root" and all the other directories say "kiva-oken". So this might be my issue and not r4ss'. 🤦 

image

e-gugliotti-NOAA:
@iantaylor-NOAA I tried your suggested fix but it didn't give everything the right permissions so I changed things back. Adding the line to change the permissions for the directory may not necessarily be needed but it might prevent other users from having the same issue that @okenk did.

iantaylor-NOAA:
Thanks for testing things out @e-gugliotti-NOAA and @okenk.
I see now that applying `chmod` to the directory is separate from changing the permissions of the files within it. Thus, 
`Sys.chmod(paths = dir, mode = "0777")` is different from `Sys.chmod(paths = list.files(dir), mode = "0777")`

However, while testing on a NMFS Mac that I have access to (but not sudo permissions on), I was able to run the status-quo `get_ss3_exe()` just fine, but running `system(paste0("sudo chmod 777 ", dir))` prompted me for my password and then resulted in `ian.taylor is not in the sudoers file.  This incident will be reported.` On the other hand, running `Sys.chmod(paths = dir, mode = "0777")` worked fine, successfully changed permissions on the directory (but not the files within it), and didn't ask for any passwords or claim to report me. I think that if I had sudo permissions, the two commands would have done the same thing.

**If our goal is to just have stuff run**
I think it would be useful to have two `Sys.chmod()` lines: one to modify the exe and one to modify the directory. Folks with more experience with linux and mac permissions can chime in with suggestions in the future if this causes any problem or isn't the right fix.

**If our goal is to respect existing file permissions**
Kiva's original error was due to not having the right permissions for the directory in which she was downloading the executable, so we could just provide an example of how to change that rather than do it automatically.

e-gugliotti-NOAA:
My goal with the function is to have stuff run. I made the change to use Sys.chmod(paths = dir, mode = "0777") in addition to changing the permissions for the exe.

iantaylor-NOAA (approved):
Looks good. Sorry for the slow response.

PR #883, opened 2023-10-16
Fix future_map call to use ... arguments to iterate_jitter

okenk:
It turns out that
```
future_map(x, ~my_func(.x, ...))
```

Is not the same as:
```
future_map(x, function(.x) my_func(.x, ...))
```

The former will not make use of the `...` arguments.

### DIFF
--- a/R/jitter.R
+++ b/R/jitter.R
@@ -169,7 +169,7 @@ jitter <- function(dir = getwd(),
     Njitter <- 1:Njitter
   }
 
-  likesaved <- furrr::future_map_dbl(Njitter, ~ iterate_jitter(
+  likesaved <- furrr::future_map_dbl(Njitter, function(.x) iterate_jitter(
     i = .x,
     dir = dir,
     printlikes = printlikes,
--- a/R/jitter.R
+++ b/R/jitter.R
@@ -161,7 +161,7 @@ jitter <- function(dir = getwd(),
   }
   r4ss::SS_writestarter(starter, overwrite = TRUE, verbose = FALSE)
 
-  # This is not necessary, but maintaining for back compatibility
+  # I'm not sure if this is necessary anymore
   file_increment(0, verbose = verbose)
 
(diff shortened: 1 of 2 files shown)


kellijohnson-NOAA:
@okenk I am curious why you are using an anonymous function here? Why not just use
```
likesaved <- furrr::future_map_dbl(
  .x = Njitter, 
  .f = iterate_jitter,
  dir = dir,
  printlikes = printlikes,
  exe = exe,
  verbose = verbose,
  init_values_src = starter[["init_values_src"]],
  ...
)
```
or will that not pass ldots through either? I was under the impression that we should only use anonymous functions when they could fit on one line and did not exist as an object.

okenk:
I am probably not doing it best. I'm pretty done solving this problem though so you are welcome to change it if it works.

iantaylor-NOAA:
Thanks @okenk for the fix and @e-gugliotti-NOAA for the test.

The revised test is passing and I tested locally after installing via `remotes::install_github("r4ss/r4ss#883")` (an option I just learned about where 883 is the PR number which results in installing the fork: `Downloading GitHub repo okenk/r4ss@main`).
I also turned on the ParmTrace and confirmed that the starting values were indeed different among the jittered models (because in the came of the `simple_small`, all 5 jitters I ran resulted in the same likelihood).

I'm going to merge this PR and then re-run the failed cmd-check github action in {nwfscDiag}: https://github.com/pfmc-assessments/nwfscDiag/actions/runs/6542080494.

iantaylor-NOAA:
Follow-up: the nwfscDiag r-cmd-check action is now able to use that packages `jitter_wrapper()` and passing the tests: https://github.com/pfmc-assessments/nwfscDiag/actions/runs/6542080494/job/17838230822#step:8:218.

chantelwetzel-noaa:
@iantaylor-NOAA Thank you for going above and beyond by running that test.

PR #884, opened 2023-10-16
Deprecate the DoProjectPlots() function

iantaylor-NOAA:
The function has been moved to https://github.com/pfmc-assessments/rebuilder as discussed in https://github.com/pfmc-assessments/rebuilder/issues/2.

This function was rarely used because it was specific to U.S. west coast groundfish stocks that were overfished and in a rebuilding plan. Therefore there's no need to have it available to all r4ss users.

This Pull Request replaces the function with a wrapper for `lifecycle::deprecate_stop()` so that Running `DoProjectPlots()` should return the following:
```
Error:
! `DoProjectPlots()` was deprecated in r4ss 1.49.1 and is now defunct.
i The function has been moved to https://github.com/pfmc-assessments/rebuilder.
```


R/RebuildPlot.R
```diff
@@ -1,635 +1,20 @@
...
-  }
-  if (pdf) dev.off()
-
-  OutputList[["ind.list"]] <- ind.list
-  return(invisible(OutputList))
+DoProjectPlots <- function(...) {
+  lifecycle::deprecate_stop(
+    when = "1.49.0",
+    what = "DoProjectPlots()",
+    details = "The function has been moved to https://github.com/pfmc-assessments/rebuilder."
```
chantelwetzel-noaa, line 18: I suggest adding to the details that the function has also been renamed to `do_project_plots`.

inst/WORDLIST
```diff
@@ -56,7 +56,6 @@ Nfleet
 Nsexes
 Nsurveys
 OFL
-Oken
```
chantelwetzel-noaa, line 59: I don't know what the wordlist does but I just wanted to confirm that this line should be removed since I see other last names in here that have been retained.

R/RebuildPlot.R
iantaylor-NOAA: @chantelwetzel-noaa, good point.

Right now both `do_project_plots()` and `DoProjectPlots()` are present in the package and @brianlangseth-NOAA is planning additional cleanup. Therefore it seems safer for now to just point users to the rebuilder repo (and thus the README). Given that the "users" are just Brian at the moment, I'm not too worried about folks getting lost.

brianlangseth-NOAA: I plan to remove do_project_plots() and keep DoProjectPlots() within the rebuilder repo. The latter was very slightly more up to date but otherwise they do the same thing. Doesn't look like that affects anything here

PR #902, opened 2024-01-18
add if statements for when release exes are names ss3_osname

e-perl-NOAA:
We discussed naming releases ss3_linux, ss3_windows, and ss3_osx. With this change for future releases, I needed to update the get_ss3_exe() function to be able to grab releases that are named either ss_whatever or ss3_whatever.


iantaylor-NOAA (approved):
Looks good. Works for me locally. We can test again after the 3.30.22.1 release.

PR #904, opened 2024-01-22
Jitter fix

### LINKED ISSUE #901: [Bug]: Jitter issues with rmarkdown
### Describe the bug

I'm trying to create a [vignette](https://github.com/PIFSCstockassessments/ss3diags/blob/FEAT-cookbook-recipe-articles/vignettes/articles/Jitter.Rmd) for ss3diags showing a jitter analysis but I am getting an error when I try to render it using the github action pkgdown.yml. I believe the issue is with running jitter() in parallel, because the backtrace from the [failed action](https://github.com/PIFSCstockassessments/ss3diags/actions/runs/7481127357/job/20362056137#step:7:121) shows: 
`  1. ├─r4ss::jitter(...)  

   2. │ └─furrr::future_map_dbl(...)  
   
   3. │   └─furrr:::furrr_map_template(...)  
   
   4. │     └─furrr:::furrr_template(...) 
    
   5. │       └─future::future(...)  
  ......`
 I was wondering if there is a way to still run jitter without the parallelization? Or if you have a suggestion for another workaround? Thanks!

### To Reproduce

1. Go to https://github.com/PIFSCstockassessments/ss3diags/actions/runs/7481127357
2. Select re-run jobs/re-run failed jobs
3. View action output

### Expected behavior

Action will fail at Deploy Package step when rendering jitter.rmd.

### Screenshots

_No response_

### Which OS are you seeing the problem on?

Linux

### Which versions of r4ss and SS3 are you seeing the problem on?

r4ss: 1.48.1, ss3: v3.30.21

### Additional Context

_No response_

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/r4ss/r4ss#code-of-conduct)

okenk:
This does two things:

- Fixes a bug where `jitter()` had some directory issues causing it to error out if it was called from a working directory other than the base model to be jittered
- Changes the way additional `...` arguments are passed to `iterate_jitter()`, per an original recommendation by @kellijohnson-NOAA that I was too lazy to implement because I thought it did not matter. Apparently it did matter.

I think this closes #901

### DIFF
--- a/R/jitter.R
+++ b/R/jitter.R
@@ -226,7 +226,7 @@ jitter <- function(dir = getwd(),
 
 #' Execute a single jittered model run
 #' 
-#' Assumes based model for the jitter is in the current working directory.
+#' Assumes the current working directory contains the base model to be jittered.
 #'
 #' @param i Index of the jitter iteration.
 #' @param  printlikes A logical value specifying if the likelihood values should
--- a/R/jitter.R
+++ b/R/jitter.R
@@ -169,21 +169,19 @@ jitter <- function(dir = getwd(),
     Njitter <- 1:Njitter
   }
 
-  likesaved <- furrr::future_map_dbl(Njitter, function(.x) {
-    iterate_jitter(
-      i = .x,
-      dir = dir,
-      printlikes = printlikes,
-      exe = exe,
-      verbose = verbose,
-      init_values_src = starter[["init_values_src"]],
-      ...
-    )
-  })
-
+  likesaved <- furrr::future_map_dbl(
+    .x = Njitter, 
+    .f = iterate_jitter,
+    printlikes = printlikes,
+    exe = exe,
+    verbose = verbose,
+    init_values_src = starter[["init_values_src"]],
+    ...
+  )
+  
   # rename output files and move them to base model directory
   to_copy <- purrr::map(Njitter, ~ list.files(
-    path = file.path(dir, paste0("jitter", .x)),
+    path = paste0("jitter", .x),
     pattern = "^[CcPRw][a-zA-Z]+\\.sso|summary\\.sso|\\.par$"
   ))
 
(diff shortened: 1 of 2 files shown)


R/jitter.R
```diff
@@ -244,15 +243,14 @@ jitter <- function(dir = getwd(),
...
                            printlikes = TRUE,
                            exe = "ss",
                            verbose = FALSE,
                            init_values_src = 0,
                            ...) {
-  jitter_dir <- file.path(dir, paste0("jitter", i))
+  jitter_dir <- paste0("jitter", i)
   copy_SS_inputs(
-    dir.old = dir, dir.new = jitter_dir, overwrite = TRUE,
+    dir.old = getwd(), dir.new = jitter_dir, overwrite = TRUE,
```
e-perl-NOAA, line 253: @okenk maybe dir should be a param, possible `dir = dir` where it uses the dir from the wrapper function and then `dir.old = dir` in case someone is trying to use this function for a model not in their current working directory?

okenk: You cannot pass it the directory from the wrapper function because that is already the working directory, set early in `jitter()`, so it would essentially look for base_jitter_dir/base_jitter_dir, which doesn't exist. That is what caused the error you and Ian were getting. That error doesn't pop up if you call `jitter()` from base_jitter_dir instead of passing it a directory argument, because then the directory is essentially empty.

This is a very non-trivial problem, but after quite a bit of trial and error, I figured out how pass it a *relative* directory, which I haven't pushed yet. Allowing the user to pass an absolute directory would require even more work that I am not up for. But I'm sure someone else could figure it out.

Let me know how you'd like to proceed.

e-perl-NOAA: Okay, that makes sense. I'm good with your changes but I'll let Ian review it as well and give his approval. Thanks for your work on this!

iantaylor-NOAA:
Sorry this has been a busy week for me with no time to look at this yet. I'll get to it as soon as I can.

iantaylor-NOAA:
This pull request was superseded by #909, closing.

PR #909, opened 2024-02-06
Try fixing jitter & file_increment functions

e-perl-NOAA:
Creating a draft PR (for now) so that it's easier to ask for input on, continue development of a fix to this issue (issue #901), and have a place for conversations about development.

@iantaylor-NOAA I altered the `file_increment()` function so that instead of renaming the files, it copies the files under a new name so that the `copy_SS_inputs()` function can find the ss3.par/ss.par file (which is where it seemed to be running into issues before).

@MOshima-PIFSC, if you would like to try grabbing this branch again and testing it out with the ss3diags jitter vignette, that would be great.

### DIFF
--- a/R/file_increment.R
+++ b/R/file_increment.R
@@ -21,7 +21,7 @@ file_increment <- function(i, verbose = FALSE,
   if (verbose) {
     message("Renaming output files to have names like Report", i, ".sso")
   }
-  ignore <- file.rename(
+  ignore <- file.copy(
     from = dir(pattern = pattern),
     to = gsub("par", "par_", gsub(
       "\\.sso|(\\.par)",
--- a/R/file_increment.R
+++ b/R/file_increment.R
@@ -1,4 +1,4 @@
-#' Rename Stock Synthesis files by adding integer value
+#' Rename key Stock Synthesis output files by adding integer value
 #'
 #' Rename files found with `pattern` by adding `i` to their
 #' name before the extension.
@@ -15,7 +15,7 @@
 #' @author Kelli F. Johnson
 #' @returns Invisibly returns a vector of logical values specifying
 #' whether or not the file was successfully renamed.
-#'
+#' @seealso [jitter()]
 file_increment <- function(i, verbose = FALSE,
                            pattern = "^[CcPRw][a-zA-Z]+\\.sso|summary\\.sso|\\.par$") {
   if (verbose) {
--- a/R/jitter.R
+++ b/R/jitter.R
@@ -258,7 +258,7 @@ iterate_jitter <- function(i,
   )
 
   # run model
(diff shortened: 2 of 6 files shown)


MOshima-PIFSC:
@e-perl-NOAA that seems to be working with `keyvec = 0:Njitter`! Thank you!

iantaylor-NOAA (approved):
Seems like this has sorted out the problem. 
Thank you @e-perl-NOAA.

I spent a long time trying to figure out why 
`future::plan(future::multisession, ...)` was crashing for me locally and `future::plan(future::sequential, ...)` was working. It turns out to be related to this issue (which I only found after figuring out the problem): https://github.com/r-lib/devtools/issues/1741.
I was relying on `devtools::load_all()` to install r4ss and did not have the most recent version installed. Apparently the package installed in that way is not present in the parallel workspaces and it uses whatever version of the package you last installed properly.

I should note that I don't understand the choice of `future_map_dbl()` vs `future_map()` in but since it works I don't care.

iantaylor-NOAA:
I should add that the `okenk:jitter-fix` branch in PR https://github.com/r4ss/r4ss/pull/904 has some additional changes that I think are beneficial so we should figure out how to combine the two things.

Sorry this has taken so long for what SHOULD be a relatively simple function. Hopefully we've all learned some useful things along the way.

kellijohnson-NOAA:
`future_map_dbl()` is guaranteed to return double values likely in a vector form where `future_map()` returns a list and the list members can be of all different classes.

okenk:
I think this looks good. I have definitely learned a lot and I hope people are not *too* annoyed that in my enthusiasm add new capabilities (I still think the capability to run jitters in parallel is quite valuable) I broke things a little.

The additional changes in PR #904 that @iantaylor-NOAA was referring to are here. I think there were some minor bugs in how the directories get handled that my PR fixed that are not included that no one had caught [here](https://github.com/r4ss/r4ss/pull/904/files#diff-63b0e69443d432a2baee832a6ada2edaf390a6ed7c923cb4080ae3bb1ec7ba19R245-R253) and [here](https://github.com/r4ss/r4ss/pull/904/files#diff-63b0e69443d432a2baee832a6ada2edaf390a6ed7c923cb4080ae3bb1ec7ba19L186-R184).

e-perl-NOAA:
Okay, thanks @okenk  I will add those! And I love that you are enthusiastic about adding new things to make the package better!

e-perl-NOAA:
I also added some messages because when I used the jitter function yesterday morning I thought it would have been useful to know which jitter I was on/where it was in the function.

iantaylor-NOAA (approved):
Looks good. Thanks @okenk and @e-perl-NOAA for all the help on this one!
I added an additional commit to run style_file() and document() to confirm that things looked good and hopefully avoid the extra PR from the github action.

Issue #924, opened 2024-03-15
[Bug]: Mohn's rho calculation 


MOshima-PIFSC:
### Describe the bug

A user pointed out that the SSB Mohn's rho values calculated by ss3diags and r4ss are inconsistent. The AFSC_Hurtado value from r4ss matches with the forecasted rho value from ss3diags (Mohn's rho for F matches). Looking deeper into the code, I believe this is because when r4ss calculates the SSB Mohn's rho, it is starting with the last year of the model but using the first retro peel estimate, whereas ss3diags starts with the first peel's last year and uses the first peel model. For example, for a model with endyr = 2022, 
r4ss method: 
`(SSB[2022, retro-1] - SSB[2022, retro0])/SSB[2022, retro0]`
vs ss3diags: 
`(SSB[2021, retro-1] - SSB[2021, retro0])/SSB[2021, retro0]`.

I believe the ss3diags method is correct because it is comparing the estimates in 2021 for the reference model (retro0) to the model with one year removed (retro-1). But r4ss is comparing the SSB estimates in 2022 for the reference model to the model with one year removed (aka the model that doesn't have data for 2022). This is also why the forecast Mohn's rho value from ss3diags matches the r4ss Mohn's rho value, because in the forecast Mohn's rho calculation, ss3diags compares SSB estimate for 2022 (one year forecast) from the retro-1 (end year 2021) model to the reference model (retro0). 

### To Reproduce

`r4ss::SSmohnsrho(retroSummary)`
`ss3diags::SSplotRetro(retroSummary)`

### Expected behavior

I expected the $AFSC_Hurtado_SSB and SSB combined Rho values to be equal. 

### Screenshots

![image](https://github.com/r4ss/r4ss/assets/78562167/92b81fd6-da35-46c9-bbb8-ce67e5d7b2ba)

### Which OS are you seeing the problem on?

Windows

### Which versions of r4ss and SS3 are you seeing the problem on?

most recent

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/r4ss/r4ss#code-of-conduct)

iantaylor-NOAA:
Fixed by #925.

PR #925, opened 2024-03-15
fix: reference year for Mohn's Rho calc #924

chantelwetzel-noaa:
This change fixes the reference year for calculating the Mohn's Rho to compare estimates to the final year with data rather than final year with data + 1 (forecast Mohn's Rho) consistently across parameters.  This should address the error found by @MOshima-PIFSC .


iantaylor-NOAA (approved):
Looks good. 
Thank you @MOshima-PIFSC for reporting the issue and @chantelwetzel-noaa for the quick fix.
This closes #924.

PR #929, opened 2024-04-02
Fix jitter dir issue

e-perl-NOAA:
Fix directory issues in the jitter function (and the file_increment function called by the jitter function) so that the function runs in parallel. There was a bug with the function running in parallel as well as if you wanted to run the function in a directory that wasn't your working directory (I was doing this trying to add a parallel option in the SS-DL-tool). 

I also added `{purrr}` and `{furrr}` to the namespaces entirely because they are generally useful packages as we try to streamline and add parallel options for functions in `{r4ss}`.


NAMESPACE
```diff
@@ -117,11 +117,12 @@ export(tune_comps)
...
 import(coda)
 import(dplyr)
+import(furrr)
 import(ggplot2)
 import(gh)
 import(magrittr)
+import(purrr)
 importFrom(corpcor,pseudoinverse)
-importFrom(furrr,future_map_dbl)
 importFrom(furrr,future_walk)
```
iantaylor-NOAA, line 126: should be able to delete this other furrr line as well

R/file_increment.R
```diff
@@ -16,16 +17,17 @@
...
     message("Renaming output files to have names like Report", i, ".sso")
   }
+
   ignore <- file.copy(
-    from = dir(pattern = pattern),
+    from = dir(path = path, pattern = pattern, full.names = TRUE),
     to = gsub("par", "par_", gsub(
       "\\.sso|(\\.par)",
-      paste0("\\1", i, ".sso"), dir(pattern = pattern)
+      paste0("\\1", 0, ".sso"), dir(path = path, pattern = pattern, full.names = TRUE)
```
iantaylor-NOAA, line 30: Since the call to `file_increment()` in `jitter()` uses `i = 0` it shouldn't be hardwired here. However, we could also just get rid of the function entirely since `jitter()` is the only place it gets used. I think the role it is playing in this parallelized version is just to backup the original files but that shouldn't be needed since the function automatically creates new separate directories.

iantaylor-NOAA (approved):
Nice work @e-perl-NOAA.
The explicit use of `dir` throughout makes a lot of sense. We should have set it up that way from the beginning.
We'll see if the code coverage passes shortly. If not, I don't think it's the fault of these changes.

R/file_increment.R
e-perl-NOAA: Ah yes, thanks for catching this. I changed it just to test locally but forgot to change it back.

iantaylor-NOAA:
Great to see the code coverage complete, @e-perl-NOAA!

We should wait until your latest commit passes (assuming it does) and then merge (or squash and merge). After that @okenk and I can rebase our fork and branch onto the updated main to get more detailed info from the code coverage workflow (assuming that your changes in 3f4cbe3 indeed provide more detail).

PR #930, opened 2024-04-04
Add pt.cex as input argument to add_legend

MOshima-PIFSC:
I added pt.cex as an argument that can be specified (instead of fixed at 0.7 as was previously) in add_legend(). This gives users more control for changing the size of points in the legend. I set the default to 0.7 so it shouldn't affect any of legends as they currently are. I also updated the documentation for the function by running `devtools::document()`.


iantaylor-NOAA (approved):
Looks good. Thank you @MOshima-PIFSC.

PR #933, opened 2024-04-08
Parallel test-test-models

e-perl-NOAA:
Update test-test-models.R to run the models and SS_output in parallel and then test the results.

### DIFF
--- a/tests/testthat/test-test-models.R
+++ b/tests/testthat/test-test-models.R
@@ -1,6 +1,6 @@
 context("Read output and make plots for all test-models")
 
-test_that("test-models work with SS_output() and SS_plots()", {
+furrr_test_that("test-models work with SS_output() and SS_plots()", {
   skip_if(!file.exists(system.file("extdata", "models", package = "r4ss")),
     message = "No 'models' folder in 'extdata'"
   )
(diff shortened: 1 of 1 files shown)


iantaylor-NOAA (changes requested):
Thanks for working on this @e-perl-NOAA.
Running the core r4ss functions on the test models in parallel could definitely speed up the github action.

It looks like the internal `run_models()` function is no longer being called which means that `SS_plots()` doesn't get called any more.

e-perl-NOAA:
Yes, clearly it's a Monday and I got too ahead of myself.

iantaylor-NOAA (approved):
Looks good, thanks for the fix.

I'm still confused by two things (neither should get in the way of merging).

First, it doesn't speed things up that much. The run-extra-tests-for-code-coverage took about 13 minutes, and previously it has taken about 17 or about 40. Looking at one of those that took about 40 minutes, it appears that 23 minutes was spent on installing R dependencies. Who knows why that happens sometimes and not others, but we may be stuck with that even with a parallel test.

Second, the code coverage metric should confirm that the test is doing all the same stuff as the old version, but for some reason, codecov isn't getting any results: https://app.codecov.io/gh/r4ss/r4ss/pulls.

e-perl-NOAA:
Hmmm, I'm not super sure about the codecov thing. When clicking on it for that particular branch I see "No Files covered by tests were changed" which makes sense because only the test themselves were changed, not the code that the tests are testing? That's the only reason I can think of so far. I'm happy to wait to merge this and see if we can find something online or ask around.

PR #939, opened 2024-05-23
fix(DESCRIPTION): Removes duplicated purrr

kellijohnson-NOAA:
There were two lines in the DESCRIPTION file with purrr as an imported package. This PR removes one of them.


iantaylor-NOAA:
Looks good. For simple fixes like this feel free to just push the change instead of Pull Request (there aren't currently branch protections in r4ss). Although I guess the PR helped me realize the error and help avoid it in the future.

Issue #927, opened 2024-03-25
get_ss3_exe() function failing for some users due to SSL error unrelated to r4ss


iantaylor-NOAA:
Thanks to @MOshima-PIFSC for pointing out that an issue that only NOAA folks (as far as I know) have been having for a few weeks with R on Windows not being able to access URLs is impacting `r4ss::get_ss3_exe()`. The function errors with the message `status was 'SSL connect error'`.

The SS3 releases are still available for manual download at https://github.com/nmfs-ost/ss3-source-code/releases/, but the convenient R function won't work for some of us until the SSL error is fixed.

If anyone has an idea for a change to r4ss that works around this, feel free to chime in. Otherwise, I'll just leave this issue up for informational purposes until the problem is resolved outside of r4ss.

iantaylor-NOAA:
Fix from @k-doering-NOAA got this working again:

> TLDR:
> 1. Run write('CURL_SSL_BACKEND=openssl', file = "~/.Renviron", append = TRUE) in your R session.
> 2. Restart your R session
> 3. Run curl::curl_version()$ssl_version and confirm the return is OpenSSL/1.1.1m (Schannel)   (note the numbers may be different)
> 4. Try remotes::install_github() (e.g., devtools::install_github('tidyverse/dplyr') ) . It should work and continue to work in future R sessions.
> 
> Full details:
> This [curl R package vignette](https://cran.r-project.org/web/packages/curl/vignettes/windows.html) explains the 2 different backends for curl for windows users and how to switch between them. My computer defaulted to secure channel (called "Schannel"), which I discovered by calling curl::curl_version()$ssl_version
> 
> I then wrote to my .Renviron file to default to using SSL instead: write('CURL_SSL_BACKEND=openssl', file = "~/.Renviron", append = TRUE)
> 
> And then restarted my r session. I double-checked the backend being used by calling curl::curl_version()$ssl_version
> The return was "OpenSSL/1.1.1m (Schannel)", confirming OpenSSL was being used (the one not in parentheses). If you have a different version of OpenSSL, the numbers may be different for you in the return.
> 
> From there, I was able to use devtools::install_github(), and utils::download.file and jsonlite::fromJSON to download files/info from github.
> 
> I think the .Renviron file will mean the backend specified will be set as the backend to use the first time curl is used in a session, so running write('CURL_SSL_BACKEND=openssl', file = "~/.Renviron", append = TRUE) one time per computer should fix the issue.
> 
[...]

PR #940, opened 2024-05-29
Add curl check for windows in get_ss3_exe()

e-perl-NOAA:
See issue #927.


e-perl-NOAA:
@kellijohnson-NOAA does this look like a good fix to you?

iantaylor-NOAA (changes requested):
@e-perl-NOAA, thanks for working on this so quickly.

I have two changes to suggest. First, I think the grepl should look for "OpenSSL" not "Schannel", because that will for sure confirm the presence/absence of the version that fixes the error for NOAA folks. The common Windows alternative is "Windows Secure Channel" (which could be theoretically abreviated "Schannel") but I can't figure out how to unset the OpenSSL setup to  figure out what the ssl_version looks like or to replicate the error when you have it set wrong.

Second, it's possible that someone would need to have a setting which isn't OpenSSL and that the function would work for them anyway (I just don't know enough about these settings to judge). So I would change `stop()` to `warning()` and add something like this to the top of the message to explain why you would want a change: 
`If this function returns "status was 'SSL connect error'", try changing the following settings:`

@kellijohnson-NOAA, I'm assuming you saw someone experience this error which inspired your comment yesterday https://github.com/r4ss/r4ss/issues/927#issuecomment-2135412436, so please feel free to suggest any changes to the above to better fit your vision for the fix.

R/get_ss3_exe.r
```diff
@@ -53,6 +53,15 @@ get_ss3_exe <- function(dir = NULL, version = NULL) {
   }
 
   if (.Platform[["OS.type"]] == "windows") {
+    if (!grepl("Schannel", curl::curl_version()$ssl_version)) {
+      stop(
+        "Please do the following:\n",
+        "1. Run write('CURL_SSL_BACKEND=openssl', file = '~/.Renviron', append = TRUE) in your R session\n",
+        "2. Restart your R session\n",
+        "3. Run curl::curl_version()$ssl_version and confirm the return is OpenSSL/1.1.1 (Schannel) (note the numbers may be different)\n",
+        "4. Try remotes::install_github() (e.g., devtools::install_github('tidyverse/dplyr')). It should work and continue to work in future R sessions."
```
kellijohnson-NOAA, line 62: I am not sure why #4 is present?

kellijohnson-NOAA:
@iantaylor-NOAA that someone was me! And, I thought to myself "oh man now I have to find the issue with the fix, I wish the instructions were just right here" but I did feel lucky that I knew where to 👀.
@e-perl-NOAA it seems as though it would be better to only provide the instructions if the function fails rather than searching for something on someone's machine given we do not know all of the cases that could lead to the failure.

iantaylor-NOAA:
@kellijohnson-NOAA, if you take away the .Renviron file, does the error return? The hard part of writing the function to only produce the message when it errors is that we need the error to test it.

kellijohnson-NOAA:
```
> r4ss::get_ss3_exe(dir="c:/users/kelli.johnson/Downloads")
trying URL 'https://github.com/nmfs-ost/ss3-source-code/releases/download/v3.30.22.1/ss3_win.exe'
trying URL 'https://github.com/nmfs-ost/ss3-source-code/releases/download/v3.30.22.1/ss_win.exe'
Error in utils::download.file(url, destfile = file.path(dir, "ss3.exe"),  : 
  cannot open URL 'https://github.com/nmfs-ost/ss3-source-code/releases/download/v3.30.22.1/ss_win.exe'
In addition: Warning message:
In utils::download.file(url, destfile = file.path(dir, "ss3.exe"),  :
  URL 'https://github.com/nmfs-ost/ss3-source-code/releases/download/v3.30.22.1/ss_win.exe': status was 'SSL connect error'
```

iantaylor-NOAA:
Great!
I can't make it come back.
We've already wrapped up `utils::download.file()` in `tryCatch()` and `suppressWarnings()`, and after playing with a few combinations of those things I can't figure out how to access the warning that gets returned by `utils::download.file()` within the function. Myabe simplest would be to just add a third step such that if it fails with "ss3" then it tries "ss" and if that also fails, then it checks the ss_version and produces the error if it's missing OpenSSL.

e-perl-NOAA:
@iantaylor-NOAA agreed and I don't get the error. Thank you Kelli for testing. I may need you to test it again once I get the changes through and working!

kellijohnson-NOAA:
I did some digging and `utils::download.file()` is using `.Internal()` which makes capturing the `stop()` difficult. But, you can capture it using the following code which I got from `demo(error.catching)`
```
tryCatch.W.E <- function(expr) {
    W <- NULL
    w.handler <- function(w) { # warning handler
      W <<- w
      invokeRestart("muffleWarning")
    }
    list(value = withCallingHandlers(tryCatch(expr, error = function(e) e),
                                   warning = w.handler),
       warning = W)
}

# Test out the function
test <- tryCatch.W.E(r4ss::get_ss3_exe(dir="c:/users/kelli.johnson/Downloads"))
> test
$value
<simpleError in utils::download.file(url, destfile = file.path(dir, "ss3.exe"),     mode = "wb"): cannot open URL 'https://github.com/nmfs-ost/ss3-source-code/releases/download/v3.30.22.1/ss_win.exe'>

$warning
<simpleWarning in utils::download.file(url, destfile = file.path(dir, "ss3.exe"),     mode = "wb"): URL 'https://github.com/nmfs-ost/ss3-source-code/releases/download/v3.30.22.1/ss_win.exe': status was 'SSL connect error'>
```

e-perl-NOAA:
@kellijohnson-NOAA I tried 2 different methods, one using the one you suggested (commit [5833c47](https://github.com/r4ss/r4ss/pull/940/commits/5833c47f1aa66b4fb15ae9cbef993ae195421b44)) and the other using a different method (commit [444c430](https://github.com/r4ss/r4ss/pull/940/commits/444c43090868d422912ba393bcbfbb4f5905f088)). Would you mind trying each of these commits to see which one works best or you like better? I have never had any OpenSSL errors on my machine so it's difficult for me to test.

kellijohnson-NOAA:
[5833c47](https://github.com/r4ss/r4ss/pull/940/commits/5833c47f1aa66b4fb15ae9cbef993ae195421b44) did not work for many reasons (1) loading the package to check the function did not work because there are two versions of the function in the R folder and the second file is used for `get_ss3_exe()` rather than the first (2) in the function there are some errors in the implementation, e.g., `try_ss` instead of `try_ss3` and `tryCatch.W.E.()` instead of `tryCatch.W.E()`. But, it appears that if you use what is returned from `tryCatch.W.E()` then things will work. 

[444c430](https://github.com/r4ss/r4ss/pull/940/commits/444c43090868d422912ba393bcbfbb4f5905f088) led to failure and the following
```
> r4ss::get_ss3_exe(dir="c:/users/kelli.johnson/Downloads")
trying URL 'https://github.com/nmfs-ost/ss3-source-code/releases/download/v3.30.22.1/ss_win.exe'
trying URL 'https://github.com/nmfs-ost/ss3-source-code/releases/download/v3.30.22.1/ss3_win.exe'
Error in utils::download.file(url, destfile = file.path(dir, "ss3.exe"),  : 
  cannot open URL 'https://github.com/nmfs-ost/ss3-source-code/releases/download/v3.30.22.1/ss3_win.exe'
In addition: Warning message:
In utils::download.file(url, destfile = file.path(dir, "ss3.exe"),  :
  URL 'https://github.com/nmfs-ost/ss3-source-code/releases/download/v3.30.22.1/ss3_win.exe': status was 'SSL connect error'
```

e-perl-NOAA:
Okay I think that I fixed it. @kellijohnson-NOAA can you try the latest commit?

kellijohnson-NOAA:
@e-perl-NOAA it partially worked. If I pass any argument, whether it is a valid directory or not, to `dir` it results in an additional error message that may or may not be true. And, it is not intuitive to the user when it says `download_location` not found when the user does not pass anything to `download_location`. `dir` is the only object name that the user actually knows. Additionally, it is still not clear to me why element (4) is in the list of instructions.
```
> r4ss::get_ss3_exe(dir="c:/users/kelli.johnson/Downloads")
trying URL 'https://github.com/nmfs-ost/ss3-source-code/releases/download/v3.30.22.1/ss3_win.exe'
trying URL 'https://github.com/nmfs-ost/ss3-source-code/releases/download/v3.30.22.1/ss_win.exe'
Error in r4ss::get_ss3_exe(dir = "c:/users/kelli.johnson/Downloads") : 
  object 'download_location' not found
In addition: Warning message:
In r4ss::get_ss3_exe(dir = "c:/users/kelli.johnson/Downloads") :
  Possible error with connect OpenSSL. Please do the following:
1. Run write('CURL_SSL_BACKEND=openssl', file = '~/.Renviron', append = TRUE) in your R session
2. Restart your R session
3. Run curl::curl_version()$ssl_version and confirm the return is OpenSSL/1.1.1 (Schannel) (note the numbers may be different)
4. Try remotes::install_github() (e.g., devtools::install_github('tidyverse/dplyr')). It should work and continue to work in future R sessions.
```

e-perl-NOAA:
Sorry, in trying to get things to work I forgot to remove that step. I pushed a different attempt to fix if you can try that out when you get a chance (given the windows check passes but it should because I ran the test on my machine).

kellijohnson-NOAA:
Now I get the following, even though nothing is downloaded.
```
> r4ss::get_ss3_exe(dir="c:/users/kelli.johnson/Downloads")
trying URL 'https://github.com/nmfs-ost/ss3-source-code/releases/download/v3.30.22.1/ss3_win.exe'
trying URL 'https://github.com/nmfs-ost/ss3-source-code/releases/download/v3.30.22.1/ss_win.exe'
The stock synthesis executable for Windows v3.30.22.1 was downloaded to: c:/users/kelli.johnson/Downloads/ss3.exe
```

e-perl-NOAA:
Ugh, okay. Just pushed what might be my last attempt before I admit defeat and might need your assistance with this one @kellijohnson-NOAA . I really wish that I could produce the error locally.

kellijohnson-NOAA:
We are back to the previous version where there is an incorrect error about the download_location
```
> r4ss::get_ss3_exe(dir="c:/users/kelli.johnson/Downloads")
trying URL 'https://github.com/nmfs-ost/ss3-source-code/releases/download/v3.30.22.1/ss3_win.exe'
trying URL 'https://github.com/nmfs-ost/ss3-source-code/releases/download/v3.30.22.1/ss_win.exe'
Error in r4ss::get_ss3_exe(dir = "c:/users/kelli.johnson/Downloads") : 
  object 'download_location' not found
In addition: Warning message:
In r4ss::get_ss3_exe(dir = "c:/users/kelli.johnson/Downloads") :
  Possible error with connect OpenSSL. Please do the following:
1. Run write('CURL_SSL_BACKEND=openssl', file = '~/.Renviron', append = TRUE) in your R session
2. Restart your R session
3. Run curl::curl_version()$ssl_version and confirm the return is OpenSSL/1.1.1 (Schannel) (note the numbers may be different)
```

iantaylor-NOAA:
I think I've sorted this out in #985, where it was easier for me to start fresh with a new branch than rebase and reuse the old one in this PR.

Issue #879, opened 2023-10-16
[Feature]: Change all instances of gender to sex


e-perl-NOAA:
### Describe the feature request.

Change gender to sex in code following issue [#516](https://github.com/nmfs-stock-synthesis/stock-synthesis/issues/516) in the stock synthesis repository. 

Instances can be found searching though [this quick search of gender](https://github.com/search?q=repo%3Ar4ss%2Fr4ss%20gender&type=code).
### Describe alternatives you have considered

Not changing. Will not be in alignment with new release of ss3.

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/r4ss/r4ss#code-of-conduct)

iantaylor-NOAA:
Thanks for posting this @e-gugliotti-NOAA. The hard part of this is described in issue https://github.com/r4ss/r4ss/issues/512, but that's not an excuse for not making progress. I will try to work on it this week.

Issue #512, opened 2021-04-06
Change column headers within r4ss read/write functions


k-doering-NOAA:
Thanks to Ian for suggesting this. I am splitting this off into a new issue.

3. I would change the column headers throughout. Even if it requires some modification to dependent code, I think the mismatch  from what's in the function inputs and the .ss_new files is confusing.

Related to number 3, it occurs to me that we could consider a refresh to all the column headers to match those used throughout data.ss_new: `yr month fleet sex part Nsamp` (from https://github.com/nmfs-stock-synthesis/stock-synthesis/blob/main/SS_write_ssnew.tpl#L191) as opposed to the `r4ss::SS_writedat()` headers, which are currently `Yr Seas FltSvy Gender Part Nsamp` (from https://github.com/r4ss/r4ss/blob/main/R/SS_readdat_3.30.R#L400)

_Originally posted by @iantaylor-NOAA in https://github.com/r4ss/r4ss/issues/505#issuecomment-814355871_

k-doering-NOAA:
Note some inputs are by month and others are by season, as noted in [this issue](https://github.com/nmfs-stock-synthesis/ss-documentation/issues/45)

iantaylor-NOAA:
Work on a scripted R workflow for adding data to the 2021 Lingcod stock assessment has highlighted for me the benefit of getting consistent terminology not just between and within r4ss and SS but also the data-processing tools like the nwfscSurvey and PacFIN.Utilities packages in use at NWFSC. The r4ss read/write functions have been vital but will be all the better when these things are more consisten.

Case also matters in R and with different column names in so many places, I ended up going with all lowercase to make it easier to remember as shown here:
https://github.com/iantaylor-NOAA/Lingcod_2021/blob/f437f9b24f010fb0367eb4206ac641ea555aef71/R/add_data.R#L400-L411
but I'd be happy with just about any convention as long as it's consistent.

k-doering-NOAA:
I was also thinking about the mismatched naming conventions within lists, @iantaylor-NOAA , while writing code for SSMSE. I like the idea of using a consistent style to make it easier to remember the names of list elements.

kellijohnson-NOAA:
I feel like it is easier on downstream users to rename things when the output and tools that work with the output all change at a consistent time. It seems like too big of a task to do all names at once, but I think it would be doable to fix a certain data set or column within all known packages at one time and work our way through. For example we could start with Year, YEAR, year, SAMPLE_YEAR.

[...]

PR #945, opened 2024-07-19
Change gender to sex

e-perl-NOAA:
Change all instances of Gender to Sex. See issue #879.

It seemed pretty straightforward but let me know if there is something amiss.

### DIFF
--- a/.github/workflows/r4ss-extra-tests.yml
+++ b/.github/workflows/r4ss-extra-tests.yml
@@ -3,11 +3,11 @@
 
 on: 
   push: 
-    branches:
-      - main
-    paths-ignore:
-      - '.github/**'
-      - 'README.md'
+    # branches:
+    #   - main
+    # paths-ignore:
+    #   - '.github/**'
+    #   - 'README.md'
   pull_request:
     branches:
       - main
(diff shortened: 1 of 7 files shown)


iantaylor-NOAA (approved):
Thanks for getting this done. I incremented the version number on main from 1.49.2 to 1.49.3 and in this branch to 1.50.0. That way if this change messes up any dependent code we can more easily confirm which version is installed.

This resolves #879.

iantaylor-NOAA:
@e-perl-NOAA, I expanded the scope of this PR to try to satisfy #512 (clean up all columns), not just #879 (gender vs sex).
Can you take a look for anything that doesn't look right?

I'm not sure if it's worth taking the time to make these changes to the functions for reading older versions of SS3 data files as those functions are now deprecated.

e-perl-NOAA:
I think it looks fine but there are some other places that I see similar column names ( like "Seas" in SS_readforecast.R) and more that might be worth changing too. Maybe we can go through the R files together in our thursday meeting to search for any more instances and figure out which terms should change and if there are any downstream impacts to other r4ss functions.  I'm happy if you want to merge this in the meantime.

PR #953, opened 2024-07-31
Fix spr series in SS_output() and get_SIS_info() to handle two table approach in SS3

e-perl-NOAA:
See SS3 PR [#609](https://github.com/nmfs-ost/ss3-source-code/pull/609) for changes in SS3 report file that needed to be changed in r4ss.

PR #954, opened 2024-08-01
Change color palette for phase plot to viridis

e-perl-NOAA:
The previous color palette was a bit confusing because it was went from hot to cold colors which made the viewer (me) think that the hot to cold colors meant something. Since they didn't I suggest using viridis for a color palette for the phase plot.


iantaylor-NOAA (approved):
Thanks for the change. 
FYI, I found info here on the use of imports via the r4ss-package.R file: https://r-pkgs.org/man.html#sec-man-package-doc.
The automated part that I was forgetting is that the use of `"_PACKAGE"` at the bottom of that file causes ?r4ss to get populated with a bunch of info from DESCRIPTION that used to be repeated in both places.

e-perl-NOAA:
@iantaylor-NOAA It still seems like there is a codecov token issue when this was merged onto main even though it passed the initial test. I'm not sure why that is though.

iantaylor-NOAA:
I reverted the change in the `spr_split_tables` branch (https://github.com/r4ss/r4ss/pull/953/commits/692fa28193009c1b0476e9c234e1167bac9adac7) so if everything else in that branch looks OK and the tests pass then we can merge on Monday and have it sorted out in main.

PR #955, opened 2024-08-02
update wordlist

e-perl-NOAA:
magrittr needed to be added


iantaylor-NOAA:
Thanks for the fix. I failed to notice that the rebased https://github.com/r4ss/r4ss/pull/954 accidentally removed the word.

Issue #950, opened 2024-07-29
speed up extra tests for code coverage


iantaylor-NOAA:
The `run-extra-tests-for-code-coverage` github action to test the code against a larger set of test models and check code coverage consistently takes 4-5 hours to run: https://github.com/r4ss/r4ss/actions/workflows/r4ss-extra-tests.yml which @e-perl-NOAA noted slows down the development process because it takes so long to learn if a change broke one of the tests or if another change fixed it.

I tried running all the tests locally on my computer and they ran much faster (about 10 minutes). Turning off the parallelization by setting `ncores <- 1` in the tests which used it increased the time to 20 minutes. I downloaded all the model files and had the exe in place so no tests were skipped:

```
══ Results ═══════════════════════════════════════
Duration: 1196.1 s

[ FAIL 0 | WARN 290 | SKIP 0 | PASS 459 ]

🌈 Your tests are over the rainbow 🌈
```

I then tried `covr::package_coverage(quiet = FALSE)` locally and it took 25 minutes to compute a coverage rate of 63.92%.

I have no idea what would be slowing down the github action. I tried revising the github action in https://github.com/r4ss/r4ss/commit/869476df7ca32dcd5338d65216f2a91bd6a8b991 which created an issue with a missing token, but I think that's been successfully added now.

My next step will be removing some of the tests to see if some subset of the tests is creating a problem.

iantaylor-NOAA:
Note, this test has often been taking more than 6 hours which times out and doesn't produce results.
I tried removing all the `furrr` commands to run parallel tests in the [sequential-tests](https://github.com/r4ss/r4ss/compare/main...sequential-tests) branch and that seemed to speed things up slightly to 4 hours and 23 minutes: https://github.com/r4ss/r4ss/actions/runs/16356907004/job/46217300510.

Unfortunately the tests are failing for a few models (it looks like in one case the model failed to run, and in two others, the plots weren't all produced). I'm not seeing these same errors when running the test locally so am really confused about what's going on.

e-perl-NOAA:
You can't tell codecov to run itself in parallel. It's on my to-do list to look into this whole issue more and figure out where things can be optimized to make the run-extra-tests-for-code-coverage any bit quicker.

PR #959, opened 2024-08-23
Bug fixes

nathanvaughan-NOAA:
These two commits correct a couple of bugs that came up working with a recent assessment due to additional options added to SS.

### DIFF
--- a/R/SS_readdat_3.30.R
+++ b/R/SS_readdat_3.30.R
@@ -607,7 +607,7 @@ SS_readdat_3.30 <-
 
     ## Size frequency methods ----
     datlist[["N_sizefreq_methods_rd"]] <- get.val(dat, ind)
-    if (datlist[["N_sizefreq_methods_rd"]] == -1) {
+    if (datlist[["N_sizefreq_methods_rd"]] <= -1) {
       # new code added for 3.30.20 to support D-M likelihood
       datlist[["N_sizefreq_methods"]] <- get.val(dat, ind)
     } else {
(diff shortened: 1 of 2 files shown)


iantaylor-NOAA (approved):
@nathanvaughan-NOAA, thanks for providing this fix. 

Could you add a comment above the new line 208 in `SS_readpar_3.30()` to explain what that line is doing because the logic is a bit convoluted (in part due to the `[[""]]` notation applied by the automatic styling).

Is it just filtering out any F parameter that don't have phase 99 (which maybe get added by the newer F method 4)?

I'm on leave all next week so leave it to @e-perl-NOAA to merge if all the tests are passing.

nathanvaughan-NOAA:
Thanks Ian will do. Yes it is just filtering out only catches for fleets that have a none 99 phase to determine how many F parameters there should be.

nathanvaughan-NOAA:
Added a descriptive comment which helped me realize this would have broken for F_method=2 so added a check that the F_4_Fleet_Parms object exists before using it to subset catch. Have a great week off, see you when you get back!

iantaylor-NOAA:
Thanks @nathanvaughan-NOAA. 
The github actions would probably have caught that issue because it runs `SS_readpar_3.30()` on a bunch of test models (in https://github.com/r4ss/r4ss/blob/main/tests/testthat/test-readwrite.R#L52-L56), but for reasons that I don't understand those extra tests take 4 hours (discussed #950), and it's a good idea to wait until they finish for a change like this, so merging will have to wait until next week after we confirm that things are working.

Issue #180, opened 2017-06-20
Figure name by SS_plots


chantelwetzel-noaa:
The current naming convention of some figures prevents their use in LaTex due to spaces or periods.   While this is not a top priority, altering all plots created by SS_plots to use a naming convention that eliminates spaces and periods would be useful for those creating documents of presentations using LaTex or other program languages.

taylori:
Good idea. I see the solutions for dealing with this at the latex end are confusing on not necessarily successful: https://tex.stackexchange.com/questions/8422/how-to-include-graphics-with-spaces-in-their-path

Looks like offending plots are "catch1 landings.png" and other catch plots with spaces, all the Francis weighting plots with periods due to "TA1.8" and anything that uses the user-supplied fleetnames if they contain spaces or periods. Fixing the first 2 would be easy in the associated functions that produce the images. Fixing the user-supplied fleetnames issue (it's nice for the captions to allow spaces in fleet names) would probably be easiest by changing the naming convention to use filenames like "index1_cpuedata_flt6.png" instead of "index1_cpuedata_NWFSC Combo Survey.png".

CWetzel:
You have found the culprits that I have encountered.  I agree that having
spaces in the user provided fleet names is ideal for document figures.  For
these figures, I have looped over a renaming function to correct the names
while retaining the nicer looking fleet names within the figures.  I think
this is not too burdensome for users to correct these themselves but it
would be nice to reduce the need for this as much as possible by correcting
the other figure names.

iantaylor-NOAA:
Fixed in PR #961 thanks to @e-perl-NOAA.
This may break some existing code if it referenced old file names like "comp_agefit_data_weighting_TA1.8_[fleetname].png", but fix should be easy to figure out (find figure in HTML view, notice that filename has an underscore instead of a period).

PR #961, opened 2024-09-05
Change figure names in plots to have better latex compatibility

e-perl-NOAA:
See issue #180
- Figure names with "." like TA1.8 changed to TA1-8
- Replace spaces in SSplotCatch figure names with underscore
- Remove spaces in SSplotIndices file names


iantaylor-NOAA (approved):
This makes sense. Thanks for the change @e-perl-NOAA.

Issue #853, opened 2023-07-20
[Feature]: better error message from `run()`


kellijohnson-NOAA:
### Describe the feature request.

When using `r4ss::run()` where the executable passed to the `exe` argument was a non-viable executable, that I obviously did not know was 🗑️, I got a non-intuitive/cryptic error message back from `system2()`, not one of the beautiful options documented in the return section of the documentation
https://github.com/r4ss/r4ss/blob/92821e749d7e72747bd010f6187b500204160128/R/run.R#L18-L20

Turns out, when a non-viable executable is passed to exe, `system2()` returns
```
error in `system2()`:
! 'CreateProcess' failed to run
'c:\github\PFMC-A~1\SABLEF~1\models\2023\base\base\test.exe`
```
I was confused because of all the tildes in the path. So, I thought maybe I had the path wrong; but, I checked the path multiple times and the file did exist. I just did not know that the executable did not work. So, I was thinking we could write a test that searches through messages captured via try/catch for the tilde and give users our own message that if the file does in fact exist that they should check that it actually works.

### Describe alternatives you have considered

Not add anything to r4ss and see if we can get system2 to return a better warning.

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/r4ss/r4ss#code-of-conduct)

iantaylor-NOAA:
@kellijohnson-NOAA, I'm happy to work on this, but I don't know how to make a 🗑️ executable like you had to replicate the issue. Can you send me something?

e-gugliotti-NOAA:
I can work on this with you Ian too.

kellijohnson-NOAA:
Sorry it has taken me so long to reply. I did not create the executable that didn't work for me, instead I downloaded it via `r4ss::get_ss3_exe()` and I am pretty sure that the executable was for a different operating system than Windows and I was on a Windows machine. I was able to recreate the error message by downloading the linux executable and trying to use it from within R after I changed the name to have the .exe extension. I got a pop up warning message when I used `system()` as well as a console message and I just got the console message when I used `system2()` which is used by `r4ss::run()`. The console message was the exact same message that I reported above. Let me know if you need any more information, and I will try to respond quicker next time 😁

iantaylor-NOAA:
@kellijohnson-NOAA, thanks for the additional info and no worries on the timeline, especially since you labeled this issue "low priority".
This problem is harder than it looks.
Running a linux executable in Windows produced a similar error for me.

[...]

PR #963, opened 2024-09-05
Better error message from `run`

e-perl-NOAA:
See issue #853

Create an error if trying to use the wrong executable.


iantaylor-NOAA (approved):
This fix looks good to me.
The situation of someone trying to run an exe from the wrong operating system is likely to be rare anyway.

PR #966, opened 2024-09-16
Add summary biomass to SSsummarize() and SSplotComparisons()

### LINKED ISSUE #524: add summary biomass to SSsummarize() and SSplotComparisons()
Comparing spawning output for models which have different maturity or fecundity relationships is problematic. Adding a comparison of summary biomass would help for these cases and may be useful in general where summary biomass can be used as a reasonable proxy for available biomass (with appropriate choice of `min age for calc of summary biomass` in the starter file).

iantaylor-NOAA:
Main changes in this PR:
- SSsummarize() now returns `$SmryBio`, `$SmryBioLower`, and `$SmryBioLower`, and `$SmryBioSD`, as well as `$accuages` (accumulator age in each model) and `$summary_ages` (the reference age used to calculate summary biomass in each model).
- SSplotComparisons() has subplot options 18 and 19 which show the summary biomass with and without uncertainty. 
  - The plots are created immediately following the biomass ratio plots, but renumbering all the plots seemed too disruptive, so I just used the next 2 numbers in the sequence.
  - If the `summary_ages` match among all models, the y-axis label will be something like "Age 3+ biomass (t)" but if they don't match it will be "Summary biomass (t)". At this point there's no warning when they don't match, but that could be easily added.
- Resolves #524.

Other minor changes folded into this PR:
- fix bug in output of `run()` which caused the results to be repeated 5 times (e.g. `[1] "ran model" "ran model" "ran model" "ran model" "ran model"`)
- add labels to the code chunks in the vignette to make debugging easier


e-perl-NOAA (approved):
Looks good!

PR #976, opened 2024-11-08
update simple_small files using ss_new from 3.30.23.00

iantaylor-NOAA:
Minor changes to file comments, headers, and formatting.


e-perl-NOAA:
Looks good to me!

iantaylor-NOAA:
This was failing for reasons that I couldn't understand and re-running the job didn't solve the problem: https://github.com/r4ss/r4ss/actions/runs/11734249018. 
I pushed an unrelated change (typo fix) which got things working again, which makes me think that the problem was related to the outdated github action and serves as a reminder that re-running the job doesn't utilize any changes in the workflow file, it just re-runs with whatever github action workflow was used the first time.

Issue #910, opened 2024-02-08
[Feature]: Surplus production vs. biomass plot


e-perl-NOAA:
### Describe the feature request.

Requested by @akatan999 in ss3 discussion [here](https://github.com/nmfs-ost/ss3-source-code/discussions/565#discussioncomment-8408231):
Following recent paper (https://cdnsciencepub.com/doi/pdf/10.1139/cjfas-2022-0168), it would be better to plot Surplus production vs. spawining stock biomass instead of total biomass. Or at least add it.

### Describe alternatives you have considered

Having users create plots themselves.

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/r4ss/r4ss#code-of-conduct)

iantaylor-NOAA:
@akatan999, see figure added via https://github.com/r4ss/r4ss/commit/6132b5308c1b8e6c2a963346b168a07b1fd85a20 to a `surplus-production` branch which you can install via `pak::pak("r4ss/r4ss@surplus-production")` (or similar with `remotes::install_github()`). Example below is for Petrale Sole showing that things look quite different vs the total biomass version.

Right now it's hard-wired to say "Spawning output" but I have an unfinished project in the works to make the spawning biomass / spawning output labels better in #838 which can apply to this plot as well.

![image](https://github.com/r4ss/r4ss/assets/4992918/ed6f6416-7353-4287-8d3a-542a85329530)

akatan999:
Thanks Ian, that was exactly my point, the two looks very different but the second is more useful from a pragmatic point of view as explained by the paper from Forrest.

iantaylor-NOAA:
Shelton Harley pointed out that the old code was wrong for seasonal models because it was summing biomass across seasons instead of taking the mean. My in-process change in the [surplus-production](https://github.com/r4ss/r4ss/compare/main...surplus-production#diff-8aba435c74cf3fe5556008d1a9ff432298fd66ad18261969ae9090bf79cee866R204-R208) branch is incorrect in the other direction because it takes the mean across seasons but also across areas which should be summed. I will try to get back to this issue before too long to make that logic correct and finish adding the alternative format with SSB instead of total biomass.

iantaylor-NOAA:
The error in the newer surplus production vs biomass plot associated with spatial models identified by Shelton Harley was resolved in Pull Request #977 thanks to @e-perl-NOAA.

PR #977, opened 2024-11-12
Resolves issue in SSplotYield.R

e-perl-NOAA:
Resolves issue #910 with most recent commit addressing the need to add across areas and calculate the mean across seasons for the surplus production plot.

### DIFF
--- a/.github/workflows/call-r-cmd-check.yml
+++ b/.github/workflows/call-r-cmd-check.yml
@@ -10,7 +10,7 @@ on:
   
 jobs:
   call-workflow:      
-    uses: nmfs-fish-tools/ghactions4r/.github/workflows/r-cmd-check.yml@r-cmd-check-attempt-fixes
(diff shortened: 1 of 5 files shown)


iantaylor-NOAA (approved):
Thanks @e-perl-NOAA for fixing this lingering issue.

I was trying to describe a remaining issue in this review when my explanation helped me realize how to solve it. In the "two_morph_seas_areas" test model, the spawning biomass summed across areas is `model$derived_quants["SSB_1971", "Value"]`
`[1] 98940`
But in the "Surplus production vs. spawning biomass plot", the line started at half that value on the x-axis. Writing that out helped me realize that the first `dplyr::summarize()` step on line 204 was replacing the NA with 0 values so it was taking the average of the SSB in season 1 with 0 in season 2 and ending up diving things by half. Removing the first `na.rm = TRUE` allowed the NAs to remain until the final step when those rows are excluded from the average.

I also added some trivial changes to the references.

PR #979, opened 2024-11-15
add token as NA to gh call in get_ss3_exe

e-perl-NOAA:
See [issue and resolution in gh package repository](https://github.com/r-lib/gh/issues/216). Resolves @efletcherPIFSC's [issue in ss3diags](https://github.com/PIFSCstockassessments/ss3diags/issues/109#issuecomment-2477541840).

### DIFF
--- a/R/get_ss3_exe.r
+++ b/R/get_ss3_exe.r
@@ -26,11 +26,11 @@
 get_ss3_exe <- function(dir = NULL, version = NULL) {
   # Get latest release if version not specified
   if (is.null(version)) {
-    latest_release <- gh::gh("GET /repos/nmfs-ost/ss3-source-code/releases/latest", page = 1)
+    latest_release <- gh::gh("GET /repos/nmfs-ost/ss3-source-code/releases/latest", page = 1, .token = NA_character_)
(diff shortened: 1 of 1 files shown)


iantaylor-NOAA (approved):
Thanks @e-perl-NOAA, @efletcherPIFSC, and all involved in finding this fix. I'm just going to merge immediately as there's nothing in the extra tests that would tell us more about this change.

efletcherPIFSC:
Will this patch be available for r4ss 1.50.0 and 1.49.3 (pre read/write function changes) releases?

iantaylor-NOAA:
@efletcherPIFSC, good point. We should issue a release/tag of 1.50.1 which includes this fix.

Making a release of the old version before the read/write changes seems convoluted. I think a cleaner solution would to copy the updated get_ss3_exe function into ss3diags until you have time to update the package to work with the current r4ss.

efletcherPIFSC:
@iantaylor-NOAA 
In the intern, until ss3diags r4ss 1.50.0+ versions, I have a R script based on the `r4ss::get_ss3_exe` with the gh token fixes

PR #980, opened 2024-11-15
Fix missing biology plots for empirical weight-at-age models

iantaylor-NOAA:
Thanks to @akatan999 for finding and reporting yet another bug in r4ss.

The change in headers to be consistent between SS3 and r4ss as discussed in https://github.com/r4ss/r4ss/issues/512 and https://github.com/nmfs-ost/ss3-source-code/pull/608 mostly impacted the r4ss functions related to reading and writing input files, but the biology plots for models that use the empirical weight-at-age depend on reading wtatage.ss_new to get info for the biology plots and I forgot to update the headers used in `SSplotBiology()` to match the new format. This implements that fix.

I also chose to make the plots of the weight-at-age information taller by default. Most png files created by `r4ss::SS_plots()` are `pheight = 6.5` inches wide and `pheight = 4` inches tall which show allow two plots per page in a document if the captions aren't too long, but a subset of them use the default `pheight_tall = 6.5` because they are better suited to a square aspect ratio, such as the phase plot.

Note that the plots of the weight-at-age info are very crude and you could do much better such as the following figure created by the code at https://github.com/pacific-hake/hake-assessment/blob/master/R/plot-heatmap-weight-at-age.R 
image

### DIFF
--- a/R/SS_plots.R
+++ b/R/SS_plots.R
@@ -483,7 +483,8 @@ SS_plots <-
         areacols = areacols,
         forecast = forecastplot, minyr = minyr, maxyr = maxyr,
         plot = !png, print = png,
-        pwidth = pwidth, pheight = pheight, punits = punits,
+        pwidth = pwidth, pheight = pheight, pheight_tall = pheight_tall,
+        punits = punits,
         ptsize = ptsize, res = res, mainTitle = mainTitle,
         cex.main = cex.main, plotdir = plotdir
       )
--- a/R/SSplotBiology.R
+++ b/R/SSplotBiology.R
@@ -385,14 +385,11 @@ SSplotBiology <-
           }
         }
       } else { ## if empirical weight-at-age IS used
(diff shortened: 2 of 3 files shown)


iantaylor-NOAA:
Thanks @e-perl-NOAA. Going through the code together would be helpful.

For the record, not specific to this issue, the challenge in cleaning all this stuff up in r4ss is that the SS3 changes in https://github.com/nmfs-ost/ss3-source-code/pull/608 did not impact the Report.sso file which still uses "Fleet", "Yr" that were changed in the input files to "fleet" and "year". I didn't try to tackle Report.sso for a few reasons:
1. The headers in the input files don't impact running the model, but changes in the output could be really disruptive to lots of peoples' custom code and other R packages.
2. The mismatch in headers between r4ss and SS3 in the input files was causing workflow headaches for me and others in addition to using incorrect terminology (#945) but the need for changes in Report.sso is less clear.

However, now there's an additional challenge of inconsistency between input and output headers which causes problems like this issue.

Issue #981, opened 2024-11-26
use updated {ghactions4r} workflows in {r4ss}


iantaylor-NOAA:
New arguments available in {ghactions4r} may require additional updates to {r4ss} as noted in https://github.com/nmfs-fish-tools/ghactions4r/issues/147:

> @e-perl-NOAA I have merged the PR. You might need to update some GHA workflows in the SS3 repositories. Let me know if you have any questions!
> 
> _Originally posted by @Bai-Li-NOAA in https://github.com/nmfs-fish-tools/ghactions4r/issues/147#issuecomment-2500889521_
>

PR #982, opened 2024-12-02
Use updated ghactions4r workflow in r4ss

e-perl-NOAA:
See issue #981.

### DIFF
--- a/.github/workflows/call-r-cmd-check.yml
+++ b/.github/workflows/call-r-cmd-check.yml
@@ -9,11 +9,11 @@ on:
   workflow_dispatch:
   
 jobs:
-  call-workflow:      
+  call-workflow:
+    env: 
+      GITHUB_PAT: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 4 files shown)


e-perl-NOAA:
This branch had some attempts to work on another issue but those were reverted and all the commits will just be squashed when merged.

e-perl-NOAA:
It's still failing because of the change to the get_ss3_exe() function to not use tokens in the gh::gh() call. With github actions, the same runner for mac is used so it calls the api enough times that the github api gets mad. I have no idea how to fix it for github actions. I tried several things but didn't get anywhere.

iantaylor-NOAA:
The fix in #986 should have this PR work fine now, but it looks like we should change the `@add-additional-args-to-yaml` to back to `@main` now that the branch has been merged in https://github.com/nmfs-fish-tools/ghactions4r/pull/147, right?

e-perl-NOAA:
Sorry @iantaylor-NOAA I didn't realize that I hadn't changed that back to main but that's why we have these reviews!

e-perl-NOAA:
It might still fail because it's using the github action from this branch, not the one from main. I would need to rebase this branch to get it passing all the checks which I can do if you would like.

PR #985, opened 2024-12-09
add useful message when get_ss3_exe() runs into "SSL connect error"

iantaylor-NOAA:
@okenk ran into the "SSL connect error" when running `r4ss::get_ss3_exe()` so I finally figured out how to set up separate R processes with and without the CURL_SSL_BACKEND set to "openssl". 

Being able to test everything side by side was definitely needed to finally get the function configured to fail with a useful message about configuring this setting only when the issue is due to "SSL connect error".

This resolves issue #927 and replaces Pull Request #940.

For future reference, among the confusing things that made this hard to solve are that the two options for `curl::curl_version()$ssl_version` on my windows computer are `"OpenSSL/3.4.0 (Schannel)"` and `"(OpenSSL/3.4.0) Schannel"`. That is, the _parentheses_ are the indicator of which option is chosen, so looking for the presence of the string "openssl" wasn't doing anything useful.


e-perl-NOAA:
This looks good to me! I just submitted a PR to fix the macos runner issue due to api calls and tokens.

iantaylor-NOAA:
The `call-doc-and-style-r` action is failing due to a bug in the ghactions4r package described here: https://github.com/nmfs-fish-tools/ghactions4r/issues/150. And the `call-r-cmd-check` is failing because the function doesn't yet have the changes that @e-perl-NOAA added to main #986.

I'm going to wait to see if there's progress on the ghactions4r issue before merging rather than apply a manual work-around which avoids the bug.

e-perl-NOAA:
Okay, it's working now it seems!

iantaylor-NOAA:
@e-perl-NOAA wise to just avoid the issue.

PR #986, opened 2024-12-10
Provide a fix to use tokens for mac github actions runner for get_ss3_exe()

e-perl-NOAA:
There are so few mac runners that it causes an error with the api call to GitHub in the `gh()` function when running r_cmd_check on GitHub actions. This fix checks if it's using a GitHub actions runner looking for the `/Users/runner/work/r4ss/r4ss` directory, which is the directory for all github actions run for `r4ss`, OR by looking for the github actions runner temp directory folder text string `/var/folders` which is what is used in the introduction vignette. If it finds either the first directory or the second text string in the directory name, then it does not provide a token parameter to the `gh()` call.

### DIFF
--- a/R/get_ss3_exe.r
+++ b/R/get_ss3_exe.r
@@ -35,7 +35,7 @@ get_ss3_exe <- function(dir = NULL, version = NULL) {
 
   # Get latest release if version not specified
   if (is.null(version)) {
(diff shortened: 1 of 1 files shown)


iantaylor-NOAA (approved):
Looks good @e-perl-NOAA, thank you for the fix.

PR #988, opened 2024-12-14
institute pause for jitter

okenk:
An [issue from the ss3 repository](https://github.com/nmfs-ost/ss3-source-code/issues/643#issuecomment-2479425005) clarified that the clock SS3 uses to set the jitter seeds is only accurate to the second. The means that, if running jitters in parallel, there is actually a very high likelihood that some seeds get reused, unlike as was originally believed. 

This implements a pause, if jitters are detected to be running in parallel, where the system pauses for `i` seconds for the `i`th jitter.

### DIFF
--- a/R/jitter.R
+++ b/R/jitter.R
@@ -60,12 +60,15 @@ SS_RunJitter <-
 #' computing, see [future::plan()]
 #'
 #' Note that random number generation occurs outside of R directly in stock synthesis.
-#' When running jitters in parallel (i.e. `future` strategy is not `sequential`), no
+#' When running jitters in parallel (i.e. `future` strategy is not `sequential`), no formal
 #' steps are taken to ensure independence of random numbers generated across
-#' cores. While the likelihood of the cores using the exact same seed is infinitesimal,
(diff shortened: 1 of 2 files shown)


iantaylor-NOAA (approved):
Thanks for adding this @okenk!

iantaylor-NOAA:
I think the doc and style workflow doesn't work for PR's from forks, so I'm going to ignore.

e-perl-NOAA:
@iantaylor-NOAA - do you think it's worth opening up an issue on ghactions4r to see if they could make that workflow work for forks?

iantaylor-NOAA:
@e-perl-NOAA, I think the issue is that we have `commit-directly: true` https://github.com/r4ss/r4ss/blob/main/.github/workflows/call-doc-and-style-r.yml#L21 but the github action can't commit to a fork. I don't know if other ghactions4r users are using commit-directly, and the best fix would probably be just skipping the github action if it detects that the PR is from a fork, which isn't so different than just doing nothing and ignoring the failed action.

PR #989, opened 2024-12-16
Solve mac api issue in get_ss3_exe() testthat

e-perl-NOAA:
Testthat was causing an error for the get_ss3_exe() function for the macos-latest runner when run more than once for a particular branch. This PR **should** fix that once and for all after figuring out an additional directory being used later in the test file for that function.

I also updated the test for the function to get rid of a warning that came up.


iantaylor-NOAA (approved):
Thanks @e-perl-NOAA!

PR #990, opened 2025-01-15
Generalize the gh command to look for any github actions runner

e-perl-NOAA:
The if statement before gh() looks for the r4ss github actions runner so that it uses a token when on that runner (this had to be done for macs). I also needed this for a new gha in the ss3-source-code repo which uses r4ss and so I made changed it to make it look for github actions runners generally instead of just looking for the r4ss runner.


iantaylor-NOAA (approved):
@e-perl-NOAA looks good. Go ahead and merge.

PR #991, opened 2025-01-29
Fix SS_output() to use URLs

kellijohnson-NOAA:
I noticed in the petrale case study in FIMS that there was a note that you could not read the report file from the URL and had to save things to your computer and create an RDS file. So, I thought I would investigate. Functions like `file.info()` and `file.exists()` do not work with a URL. This PR could be made better by using a function to check if the file exists that incorporates URL and file checks rather than rewriting the code over and over.

Note that, this was not fixed for the log file because I do not know what to do when there are multiple files. This branch works for petrale now in the FIMS case study.

### DIFF
--- a/R/SS_output.R
+++ b/R/SS_output.R
@@ -212,7 +212,7 @@ SS_output <-
     if (!is.character(dir) | length(dir) != 1) {
       stop("Input 'dir' should be a character string for a directory")
     }
-
+browser()
     # get info on output files created by Stock Synthesis
     shortrepfile <- repfile
     repfile <- file.path(dir, repfile)
@@ -236,7 +236,19 @@ SS_output <-
         stop("report file is empty: ", repfile)
       }
     } else {
-      stop("can't find report file: ", repfile)
+      if (grepl("https:", repfile)) {
+        con <- url(repfile)
+        check <- suppressWarnings(try(
+          open.connection(con, open = "rt", timeout = 2),
+          silent = TRUE
+        )[1])
+        suppressWarnings(try(close.connection(con), silent = TRUE))
+        if (!is.null(check)) {
+          stop("can't find report file: ", repfile)
+        }
+      } else {
+        stop("can't find report file: ", repfile)
+      }
     }
     rephead <- readLines(con = repfile, n = 50)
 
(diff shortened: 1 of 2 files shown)


e-perl-NOAA:
Running the following returned results for me! Thanks for working on this!
```
petrale_output <- r4ss::SS_output("https://raw.githubusercontent.com/pfmc-assessments/petrale/main/models/2023.a034.001/")
```

kellijohnson-NOAA:
Thanks @e-perl-NOAA for removing the browser() call for me, sorry about that.

kellijohnson-NOAA:
Looks like there are still some other things to chase down that I screwed up on. I should have done this as a draft PR or an issue 🤦 with a proposed fix. I was just excited that I found something that was slightly working. So @iantaylor-NOAA and @e-perl-NOAA feel free to provide guidance on how you would like further development to go on this. Happy to help but also willing to step back and let someone else take it across the finish line.

e-perl-NOAA:
@kellijohnson-NOAA No worries! I do that all the time haha. It seems like something is going awry in the vignette so I'm running it locally to see if I can figure out what that might be.

kellijohnson-NOAA:
I am guessing that I implemented something wrong in the the nested if else statements with respect to the compfile.

iantaylor-NOAA:
@kellijohnson-NOAA, this is awesome. Thank you for investigating and finding a solution.
Coincidentally I was just looking at the code for reading the log files today and noted that it's messy. The only goal for reading the ss3.log file (or whatever it's called if your exe has a different name) is to check for temporary files being written which slows down estimation. If you've already uploaded your model output to github, then you've probably taken the time to run `SS_output()` locally and deal with any issue found in the log file, so we could just turn off the read of the log file when reading from the http addresses.

e-perl-NOAA:
I think I found what's going wrong in the function causing the error.

e-perl-NOAA:
@kellijohnson-NOAA I was able to fix what was going wrong in the code if you want to give it a final look through.

iantaylor-NOAA:
@kellijohnson-NOAA  and @e-perl-NOAA thanks for all your work on this new functionality. There's nothing in the extra tests that should fail related to this issue, so I'm happy to merge today rather than wait for that test to finish. Let me know if there's anything remaining to be done before or concerns about merging now.

Issue #995, opened 2025-03-01
[Bug]: Inconsistencies delimiting "Opening Connection to " message with file. (SS_writestarter)


efletcherPIFSC:
### Describe the bug

There is no space in between the message "Opening Connection to" and file path when this message is invoked in SS_writestrarter`. 

There are similar inconsistencies with   
- https://github.com/r4ss/r4ss/blob/b2b441825b4d4c0465c4b85b94c1d34a0e463f50/R/SS_varadjust.R#L202
- https://github.com/r4ss/r4ss/blob/b2b441825b4d4c0465c4b85b94c1d34a0e463f50/R/SS_writeforecast.R#L52
- https://github.com/r4ss/r4ss/blob/b2b441825b4d4c0465c4b85b94c1d34a0e463f50/R/SS_writewtatage.R#L55
- https://github.com/r4ss/r4ss/blob/b2b441825b4d4c0465c4b85b94c1d34a0e463f50/R/SS_writectl_3.24.R#L50

### To Reproduce

1. Copy and paste a *starter.ss* to a test directory. Set working directory to that directory.
2. Read In *starter.ss* file via `r4ss::SS_readstarter` save to object x  
    - `x <- r4ss::SS_readstarter()`
3. Pass object `x` to `r4ss::SS_writestarter()`, to see messages. To reproduce, output directory to `tempdir()`, save to file *new_starter.ss* and `overwrite=TRUE` 
    - `r4ss::SS_writestarter(x, dir = tempdir(), file="new_starter.ss", overwrite=TRUE)`

### Expected behavior

Spacing between the SS_writestarter message output and file path. 

Reproducing behavior will produce something similar to the following output:

```
> SS_writestarter(x, dir=tempdir(), file="new_starter.ss", overwrite=TRUE)
running SS_writestarter
writing new file:C:\Users\[username]\AppData\Local\Temp\RtmpIzHcaj/new_starter.ss
opening connection toC:\Users\[username]\AppData\Local\Temp\RtmpIzHcaj/new_starter.ss
file written to C:\Users\[username]\AppData\Local\Temp\RtmpIzHcaj/new_starter.ss
```

### Which OS are you seeing the problem on?

Windows

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/r4ss/r4ss#code-of-conduct)

e-perl-NOAA:
[...]

PR #997, opened 2025-03-03
Fix open connection spacing in SS_writestarter

e-perl-NOAA:
See issue #995 to fix the spacing in the console output of SS_writestarter().

### DIFF
--- a/R/SS_varadjust.R
+++ b/R/SS_varadjust.R
@@ -199,7 +199,7 @@ SS_varadjust <- function(dir = "C:/myfiles/mymodels/myrun/",
   }
 
   # open connection to file
-  if (verbose) message("opening connection to", newctlfile)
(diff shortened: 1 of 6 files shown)


iantaylor-NOAA (changes requested):
Thanks @e-perl-NOAA. The changes to `SS_writestarter()` look good but I think we should make similar changes in the other files that @efletcherPIFSC identified in #995. Additionally, we could remove the "\n" in https://github.com/r4ss/r4ss/blob/b2b441825b4d4c0465c4b85b94c1d34a0e463f50/R/SS_writewtatage.R#L55 so it matches the others.

iantaylor-NOAA (approved):
Thanks @e-perl-NOAA, sorry for the slow review.

Issue #996, opened 2025-03-03
[Bug]: SSsummarize issue with areas


e-perl-NOAA:
### Describe the bug

Carey McGilliard sent an email to Kathryn and then me detailing the following:
I'm having with the r4ss::SSsummarize function for my 2-area model. I get the error message:

```
Error in [<-.data.frame(*tmp*, SmryBio[["Yr"]] %in% NA_yrs, imodel, :
replacement has 88 rows, data has 44
```
I can run SSsummarize for a one-area model (as can others), and for a one-area two-sex model. The fact that there are half the number of rows in the data makes me think that an update is not compatible with the 2-area model.

### To Reproduce

1. Run the three_area_nomove model from the ss3-test-models repo
2. biglist <- SSgetoutput(dirvec = dir)
3. SSsummarize(biglist)

### Expected behavior

The function would not produce an error and would be able to handle multiple areas.

### Which OS are you seeing the problem on?

Windows

### Which versions of r4ss and SS3 are you seeing the problem on?

most recent versions of both

### Additional Context
@iantaylor-NOAA 

Looking in the SSsummarize() function, the error is happening here:
https://github.com/r4ss/r4ss/blob/b2b441825b4d4c0465c4b85b94c1d34a0e463f50/R/SSsummarize.R#L623-L625

Adding `& Area == 1` after `Seas == 1` made the error go away. However, I don't know if we necessarily want to only have SmryBio info only for area 1 in the summary. If we do need all 3 areas, I think that the new function really should just pull the SmryBio data from the timeseries rather than derived quantities.

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/r4ss/r4ss#code-of-conduct)

PR #998, opened 2025-03-07
Fix to SSsummarize

e-perl-NOAA:
Fixes issue #996 by combining areas and sexes to get the SmryBio output. This now works and provides the sum(SmryBio) that we expect from the time_series table in the report file, though you might want to test it a bit more than I can right now, so I have included this as a draft PR.

### DIFF
--- a/R/SSsummarize.R
+++ b/R/SSsummarize.R
@@ -621,7 +621,9 @@ SSsummarize <- function(biglist,
     if (length(NA_yrs) > 0) {
       # filter years to only include those with NA values
       SmryBio[SmryBio[["Yr"]] %in% NA_yrs, imodel] <- ts |>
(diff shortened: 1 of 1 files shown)


iantaylor-NOAA:
Looks good!
I copied the example input from https://nmfs-ost.github.io/ss3-doc/SS330_User_Manual_release.html#controls-for-variance-of-derived-quantities into the "read specs for more stddev reporting" of a spatial model and compared the results with and without this change in the plot below. In this case the red line is the summary biomass from the derived quantities with associated uncertainty while the matching blue line is the summary biomass from the time series table aggregated across areas thanks to the change that @e-perl-NOAA just made.
image

iantaylor-NOAA:
I forgot to give credit to @akatan999 for doing additional testing on this fix.

PR #1003, opened 2025-04-16
fix: add discard parameters for age-based model

chantelwetzel-noaa:
I found that the `SS_readctl_3.30()` was not reading the control file correctly for models with age-based discarding and retention.  I have not adding a similar fix to the `SS_readctl_3.24()` function since I don't think that is necessary, plus I don't think age-based retention was correctly implemented in 3.24 models.


iantaylor-NOAA (approved):
Thanks @chantelwetzel-noaa. The code changes look right to me. I haven't tested this on a model with age-based retention, but am assuming you've done so and it works.

Also, I'm assuming that `SS_writectl_3.30()` works fine such that you can run
```
x = SS_read(dir1)
SS_write(x, dir2)
y = SS_read(dir2)
```
and have `x` and `y` match, but I'm sure your work with the revised function will be enough to test it thoroughly.

PR #1005, opened 2025-04-30
Would comp2long() be a useful function?

arni-magnusson:
Hi r4ss team!

The SPC team is currently working with length and weight composition data in Stock Synthesis and I wanted to access the data, explore, and analyze. For exploration and analysis, one would usually convert the data to long format. Would it be useful to have a function in r4ss to allow users to perform this task with a short single line?

The examples on the comp2long() help page describe what the function does. It also works with weight comps (and other generalized size comps), although that's not demonstrated in the examples.

I'm curious to hear what you think. Feel free to reject, rename, modify, or propose any changes. Implemented in base R, this function should require little or no maintenance.


iantaylor-NOAA:
@arni-magnusson, thanks for the contribution! This seems really useful for exploring the comp data. Note that the comp data output is already in long format but it's nice to be able to explore the data without running through the model first.

Since the examples starting in https://github.com/r4ss/r4ss/blob/7443dd5084754bf6ce475790b2050b54a4774920/R/comp2long.R#L34 are not wrapped in `\dontrun{}` they will be run via the call-r-cmd-check workflow. That's not quite as good as having a testthat test to confirm that the answers look right, but I think is fine for now.

PR #1006, opened 2025-04-30
ghactions4r updates

e-perl-NOAA:
ghactions4r has changed some of their workflows (the calc-cov one), added a workflow to create a badge showing coverage, and will be moving the repo to the nmfs-ost org on May 16 (see discussion [here](https://github.com/nmfs-fish-tools/ghactions4r/issues/159#issuecomment-2840460572))

As Kelli notes below, there is a fork on nmfs-ost already so this can be merged.


iantaylor-NOAA:
@e-perl-NOAA, thanks for keeping us up to date with changes in ghactions4r!

PR #1010, opened 2025-05-07
Improving comp2long()

arni-magnusson:
Hi! Thanks for adding `comp2long()` to the package. This pull request adds the following improvements to the function:

- comp2long now handles male and female composition data better - the earlier version did not distinguish between the sexes
- comp2long now supports age composition data in addition to length and size comps - the earlier version did not convert age composition data
- comp2long is now a wrapper function that calls either `age2long` or `size2long` to perform the conversion, depending on the data type of `x` - the user can also call the lower-level functions directly, but the examples demonstrate that comp2long is convenient and does the job

As before, feel free to rename, modify, or propose any changes.


R/comp2long.R
```diff
@@ -1,31 +1,37 @@
...
-#' for length composition data, or
-#' \preformatted{method  year  month  fleet  sex  part  Nsamp  size  freq}
-#' for generalized size composition data.
+#' If \code{x} is not a data frame, the function will look for \code{lencomp}
+#' inside the list.
 #'
-#' The second to last column can be named explicitly via the measure argument.
+#' The \code{expand = TRUE} option is mainly useful for accessing conditional
+#' age-at-length, e.g., to produce a data frame that has the same number of rows
+#' as the number of otoliths.
```
iantaylor-NOAA, line 26: `otoliths` isn't in WORDLIST. You can update it via `spelling::update_wordlist()`

iantaylor-NOAA (approved):
@arni-magnusson, thanks for expanding these functions. At this point you are the only user of them so I'm happy with whatever level of refinement you feel like providing. The spell check fix to get the github actions passing is the only change that would be really valuable at this stage.

arni-magnusson:
Thanks for the review and good feedback. I've added otoliths to the WORDLIST and improved the data frame search algorithm. As the updated comp2long() help page says:
> If `x` is not a data frame, the function will look for a data frame inside the list: first `lencomp`, then `agecomp`, and then `sizefreq_data_list[[1]]`.

The function now handles the two cases that you mention: the empirical weight-at-age model and ```inputs$dat$sizefreq_data_list[[1]] |> comp2long()```

iantaylor-NOAA:
Looks great, thank you @arni-magnusson!
Sorry I should have had you on the author list already, but it's all the more obvious now.

PR #1017, opened 2025-05-29
fix SS_decision_table_stuff() for spatial models and improve SS_ForeCatch() for those with mixed catch units

iantaylor-NOAA:
The `SS_decision_table_stuff()` function has been completely rewritten to use dplyr and aggregate across areas.
The `SS_ForeCatch()` function has been updated to provide a message when your units of catch are mixed between numbers and biomass and to clarify in the comment on the resulting output that the sums are in mixed units. This doesn't change the numerical values in the table, just clarifies the comment to help with confusion like what I just went through.

Tagging @rclairer.


R/SS_decision_table_stuff.R
```diff
@@ -11,35 +11,36 @@
...
-  SpawnBioVirg <-
-    replist[["timeseries"]][["SpawnBio"]][replist[["timeseries"]][["Era"]] == "VIRG"]
-  dep <- round(SpawnBio / SpawnBioVirg, digits[3])
-  # get summary biomass (not currently reported)
-  Bio_smry <- ts[["Bio_smry"]]
-  # combine stuff
-  # stuff <- data.frame(yr=yr[ts[["Area"]]==1], catch, dep, SpawnBio, Bio_smry)
-  stuff <- data.frame(yr, catch, SpawnBio, dep)
-  return(stuff)
+SS_decision_table_stuff <- function(replist, yrs = 2025:2036, digits = c(0, 0, 3)) {
```
e-perl-NOAA, line 21: I know I'm a little late to the party and I can put this in a separate issue/work on it myself, but instead of having to update this function every assessment cycle with the correct years, should we just use the datetime functionality to get the current year and add 11 years to that?

PR #1022, opened 2025-06-04
Fix ss_readctl_3.30 q offset

### LINKED ISSUE #1020: [Bug]: `SS_readctl()` not reading correctly when catchability link = 5 (offset) is used
### Describe the bug

Catchability link option 5 which is recommended with survey types 35 and 36 requires an offset parameter after the catchability parameter line. The `SS_readctl()` function is not working correctly with this link option because it is not expecting the offset parameter line which results in an error when the selectivity set-up section is read.  I have not checked `SS_writectl()` but I suspect there may be a similar issue with this function as well.

### To Reproduce

1. Add a survey that is either survey type 35 or 36 in the data file.
2. Set the link type to 5 in the catchability set-up table in the control file.
3. Include a catchability and offset parameter line in the catchability parameter lines for this survey.

### Expected behavior

`SS_readctl()` should read the link type and then expect to read an offset parameter line before moving on to the selectivity set-up lines.

### Screenshots

_No response_

### Which OS are you seeing the problem on?

_No response_

### Which versions of r4ss and SS3 are you seeing the problem on?

_No response_

### Additional Context

_No response_

e-perl-NOAA:
Adds a section in the SS_readctl_3.30.r file to add a parameter line when there is a link = 5. I also added a section that when the link = 5, it uses the text "Q_base" instead of "LnQ_base" as this is the format that the manual has.

Closes #1020

### DIFF
--- a/R/SS_readctl_3.30.R
+++ b/R/SS_readctl_3.30.R
@@ -895,22 +895,21 @@ SS_readctl_3.30 <- function(file, verbose = FALSE,
     {
       if ((ctllist[["Q_options"]][j, ][["float"]] == 0) || (ctllist[["Q_options"]][j, ][["float"]] == 1)) # handle float 0 or 1 as 1 parm sel
         {
-        flname <- fleetnames[ctllist[["Q_options"]][j, ][["fleet"]]]
-         if(ctllist[["Q_options"]][j, ][["link"]] == 5)
-         {
-          comments_Q_type[[i]] <- paste0("Q_base_", flname, "(",
-            ctllist[["Q_options"]][j, ][["fleet"]], ")",
-            collapse = ""
-          )
-         } else {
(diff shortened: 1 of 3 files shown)


iantaylor-NOAA (approved):
Looks good. Thanks @e-perl-NOAA!

iantaylor-NOAA:
Let's wait until after the extra tests are complete and then merge.

iantaylor-NOAA:
@chantelwetzel-noaa, sorry, I'm looking at this test, which started 5 hours ago but takes forever but seems to never take longer than 6 hours): https://github.com/r4ss/r4ss/actions/workflows/r4ss-extra-tests.yml

chantelwetzel-noaa:
Sorry, I did not see that test still running! I agree that we should wait.  Thank you.

Issue #941, opened 2024-06-10
[Feature]/[BUG]: SS3 expanded output for SPAWN_RECR output table (SS3 added a,b formulation) 


e-perl-NOAA:
### Describe the feature request.

Add capability of `SS_output()` to read the expanded output of the SPAWN_RECR output table (see ss3-source-code issue [#191](https://github.com/nmfs-ost/ss3-source-code/issues/191#issuecomment-2159045859)).

In attempting the [example that Rick attached to the PR](https://github.com/user-attachments/files/20581465/TimeVary_growth.zip), it looks like r4ss isn't reading the output correctly given the changes and there are downstream impacts causing errors in the `SS_fitbiasramp()` and `SS_plots()` functions.

Also update SS_read() with changes to input files:
-  [x] Control file: b-h_ab option 10 in spawner recruit params
   ```
   "SR_LN(R0)_derived"
   "SR_BH_steep_derived";
   "SR_BH_ln(alpha)";
   "SR_BH_ln(beta)";
   ```
- [x] Starter file:
     - [x] 1 #_Compatibility:  flag for legacy (0) vs improved (1) impact of timevary biology on benchmark SRR calcs >=3.30.24
- [x] Forecast file (last line in the following):
   ```
   1 # Control rule method (0: none; 1: ramp does catch=f(SSB), buffer on F; 2: ramp does F=f(SSB), buffer on F; 3: ramp does catch=f(SSB), buffer on catch; 4: ramp does F=f(SSB), buffer on catch) 
   # values for top, bottom and buffer exist, but not used when Policy=0
   0.4 # Control rule inflection for constant F (as frac of Bzero, e.g. 0.40); must be > control rule cutoff, or set to -1 to use Bmsy/SSB_unf 
   # Also see HCR_anchor below
   0.1 # Control rule cutoff for no F (as frac of Bzero, e.g. 0.10) 
   0.75 # Buffer:  enter Control rule target as fraction of Flimit (e.g. 0.75), negative value invokes list of [year, scalar] with filling from year to YrMax 
   #
   3 #_N forecast loops (1=OFL only; 2=ABC; 3=get F from forecast ABC catch with allocations applied)
   3 # First forecast loop with stochastic recruitment
   1 # Forecast base recruitment:  0= spawn_recr; 1=mult*spawn_recr_fxn; 2=mult*VirginRecr; 3=deprecated; 4=mult*mean_over_yr_range
[...]

PR #1025, opened 2025-06-25
Fix read and write files for t-v biology and S-R curve a,b option

e-perl-NOAA:
- updated read and write ctl file
- updated read and write starter file
- updated read and write forecast file

See checklist in issue #941

### DIFF
--- a/R/SS_readforecast.R
+++ b/R/SS_readforecast.R
@@ -14,7 +14,7 @@
 #' @author Ian G. Taylor, Kelli F. Johnson, Kathryn L. Doering, Nathan R. Vaughan
 #' @export
 #' @family read/write functions
-file <- "C:\\Users\\elizabeth.gugliotti\\Desktop\\timevary_growth_files\\forecast.ss_new"
+
 SS_readforecast <- function(
   file = "forecast.ss",
   Nfleets = NULL,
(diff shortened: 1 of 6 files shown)


R/SS_readforecast.R
```diff
@@ -14,7 +14,7 @@
 #' @author Ian G. Taylor, Kelli F. Johnson, Kathryn L. Doering, Nathan R. Vaughan
 #' @export
 #' @family read/write functions
-
+file <- "C:\\Users\\elizabeth.gugliotti\\Desktop\\timevary_growth_files\\forecast.ss_new"
```
iantaylor-NOAA, line 17: added by accident

R/SS_writestarter.R
```diff
@@ -79,6 +79,11 @@ SS_writestarter <- function(
   # writes the content of an R object, followed by the object name with "#_" in front
   wl <- function(name) {
     value <- mylist[names(mylist) == name]
+    if (is.character(value) && grepl("^c\\(.*\\)$", value)) {
```
iantaylor-NOAA, line 82: could you add a comment explaining what the regular expression is doing (copilot is good at this but nicer to have available for everybody)?

iantaylor-NOAA (approved):
Thanks @e-perl-NOAA! 
I didn't run the updated functions on the example model with the new stock-recruit option, but obviously you did as part of the development.

There line left in by accident in SS_readforecast.R is being misinterpreted as an additional element in the read/write family of funcitons causing all the .Rd file changes. Removing that and re-running `devtools::document()` should remove the changes to many of those files.

PR #1030, opened 2025-07-03
Catch plots 1&2 label fix for models with retention

### LINKED ISSUE #1029: [Feature]: Revise the y-axis label of the catch1_landings.png and catch2_landings_stacked.png
### Describe the feature request.

The default y-axis labeling for the catch1_landings.png and catch2_landings.stacked.png figures is Landings (t) for both models that do or do not estimate retention. This is only correct for models that estimate retention where `r4ss` creates additional figures that show total catch (landings + estimated discard mortality).  For models that do not include retention, these two figures should by labeled as "Total catch (t)" on the y-axis. 

### Describe alternatives you have considered

None

### Statistical validity, if applicable

_No response_

### Describe if this is needed for a management application

_No response_

### Additional context

_No response_

e-perl-NOAA:
Resolves #1029 by switching between labels if there are retention parameters.


chantelwetzel-noaa:
While all the plots are now present, this is still incorrect.  If a model includes retention the y-axis label for catch1_landings.png and catch2_landings stacked.png should be "Landings (t)" and  for a model that does not include retention, the y-axis label should be "Total catch (t)".  This is the opposite of what the code is currently doing.

chantelwetzel-noaa:
Let me know when you have a fix that you would like me to test out.

e-perl-NOAA:
@chantelwetzel-noaa Please try it now, I think I have it working

chantelwetzel-noaa:
This still is not correct. The label is correct for models with retention, but the y-axis label for a model without retention is still "Landings (t)" where it should be "Total catch."

e-perl-NOAA:
Ugh, okay. This is harder to fix than one would think for some reason.

PR #1038, opened 2025-07-24
"fix SS_changepars() input string matching"

kinmic:
Adjusted `SS_changepars()` function to check whether `strings` input is an exact match to a parameter name in `control.ss_new`. If it is, the matching parameter will be modified by the function. If no exact match is found, the the function will check for parameter names which partially match `strings` and modify these.

### DIFF
--- a/R/SS_changepars.R
+++ b/R/SS_changepars.R
@@ -17,7 +17,10 @@
 #'   parameters to be modified. This is an alternative to `linenums`.
 #'   `strings` correspond to the commented parameter names included in
 #'   `control.ss_new`, or whatever is written as comment at the end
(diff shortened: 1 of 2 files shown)


kinmic:
call-doc-and-style-r.yml workflow is failing as the .yml is designed to work on main on not a branch. Would you like  me to resubmit the PR from a main in my fork and not use a branch?

iantaylor-NOAA (approved):
Changes all look good to me.
I've just been ignoring the doc and style workflow issue (which can be sorted out separately), especially in cases like this where you've already run `devtools::document()` and the styling looks fine.

PR #1039, opened 2025-07-25
Switch to sequential tests for extra testing to calculate code coverage

### LINKED ISSUE #950: speed up extra tests for code coverage

iantaylor-NOAA:
This test rewrites the test-test-models.R script called by the r4ss-extra-tests.yml workflow with two goals:
1. speed up the code coverage calculations conducted by codecov. It thus fixed #950. I have no idea why, but running the code on models in sequence instead of parallel (with {furrr}) improved the speed from > 6 hours to about 4.5 hours.
2. provide more informative messages about which model is failing when there's a failure. Because the tests are run in sequence, I could put the call to `test_that()` inside the loop and have the test description include the model name. The resulting message below (copied from https://github.com/r4ss/r4ss/actions/runs/16503548011/job/46668255127#step:12:535) made it easy to diagnose that the problem was caused by running the test-models with the "latest" SS3 release (version 3.30.23.2), on a test model that had been modified to include an extra input line required to test a feature included in the 3.30.24 prerelease.

```
══ Failed tests ════════════════════════════════════════════════════════════════
── Error ('test-test-models.R:49:7'): run SS_output() and SS_plots() on test model: growth_timevary ──
Error in `eval(code, test_env)`: object 'output' not found
```

Tagging @k-doering-NOAA and @Bai-Li-NOAA in case you find this interesting since you both are involved in test configurations.

### DIFF
--- a/.github/workflows/r4ss-extra-tests.yml
+++ b/.github/workflows/r4ss-extra-tests.yml
@@ -70,7 +70,7 @@ jobs:
 
       - name: Get the latest SS3 executable and move to expected location
         run: |
-          wget -O ss3 https://github.com/nmfs-ost/ss3-source-code/releases/latest/download/ss3_linux
+          wget -O ss3 https://github.com/nmfs-ost/ss3-source-code/releases/download/v3.30.24-prerel/ss3_linux
           sudo chmod a+x ss3
           cp ss3 inst/extdata/simple_small/ss3
           rm ss3
--- a/tests/testthat/test-test-models.R
+++ b/tests/testthat/test-test-models.R
@@ -1,83 +1,82 @@
 context("Read output and make plots for all test-models")
 
-test_that("test-models work with SS_output() and SS_plots()", {
-  skip_if(
-    !file.exists(system.file("extdata", "models", package = "r4ss")),
-    message = "No 'models' folder in 'extdata'"
-  )
-  # skip if no executable in simple_small path
-  # (should have been loaded there by
-  # .github\workflows\r4ss-extra-tests.yml)
+# temporary directory
+mod_path <- file.path(tempdir(check = TRUE), "test-test-models")
+download_models(dir = mod_path)
+mods <- list.dirs(
+  file.path(mod_path, "models"),
+  full.names = FALSE,
+  recursive = FALSE
+)
 
-  # find simple_small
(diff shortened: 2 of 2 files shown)


iantaylor-NOAA:
@Bai-Li-NOAA, you're right that there's lots of redundancy, but as @e-perl-NOAA notes, different features are included in different models. Looking back at a comment I made when this set of tests was added, https://github.com/r4ss/r4ss/issues/316#issuecomment-1590160846, I noted "The change bumps the coverage up from 52% to 59% since the simple_small model doesn't have as many data types as the larger set of models so plotting functions like SSplotDiscard() were shown with 0% coverage (now it's at 85%). "

These tests don't get triggered by call-r-cmd-check because the workflow doesn't download the models required to do all the extra testing. It's theoretically possible that we could do more targeting testing by skipping the plots that occur in all models, but I'm not sure how much that would speed things up. Obviously faster is nicer but as long as it runs in less than 6 hours and finishes without timing out, I'm OK with that.

e-perl-NOAA (approved):

PR #1043, opened 2025-08-14
Improve support for two-sex format in comp2long()

arni-magnusson:
My pull request from 2 days ago was an improvement, but it turns out it was still not a reliable approach for detecting whether the input composition data frame was in two-sex side-by-side format or not.

With this pull request, instead of using sex = 0, 1, 2, or 3 to determine whether the comp data are in two-sex format, I simply read the column names and check if bin names are repeated (for each sex) or not.


iantaylor-NOAA:
Looks good. Thanks @arni-magnusson.

Issue #1050, opened 2025-08-28
[Feature]: Better naming of `SSgetoutput()` and `SSsummarize()`


okenk:
### Describe the feature request.

`SSgetoutput()` and `SSsumarize()` are very useful for collecting results from a suite of models, but they do not actually track what models they contain because the name of each model must be `replist1`, `replist2`, etc. If the models are going to be named, they should accept descriptive names that can help the user remember what is in each element.

That is, `SSgetoutput()` should accept a vector of model names that will replace `replist1`, etc. as the names of the elements of the list. Currently the user can manually set these names after reading in the models, which is not too hard.

However, those descriptive names are lost when `SSsummarize()` summarizes the list of models, and it reverts back to `replist1`, etc., and those names are not easy to set manually. This could make it so users do not need to set the names every time they call table comparisons and plot comparisons, instead they get pulled from a useful name.

I think this will make for better organization and less likelihood of accidentally mislabeling things.

### Describe alternatives you have considered

Continue to use `replist1`

PR #1051, opened 2025-08-28
Use modelnames throughout the comparison functions

iantaylor-NOAA:
As suggested by @okenk in #1050, this improves functionality of the model comparisons functions by 
1. adding a `modelnames` argument to `SSgetoutput()`,
2. that argument is used by `SSsummarize()` and now passed from there to `SStableComparisons() and `SSplotComparisons()` (where it is used as the default for `legendlabels`)

Other minor changes
* add example to documentation of `SSgetoutput()`
* improve internal links between documentation of all comparison functions
* remove long-ago deprecated argument to `SSgetoutput()` and newly deprecate an non-functioning argument

@okenk, it would be great if you could check if these change works the way you were hoping they would

### DIFF
--- a/R/SSgetoutput.R
+++ b/R/SSgetoutput.R
@@ -15,9 +15,7 @@
 #' @param forecast Choice to read or not read forecast quantities.
 #' Default=FALSE.
 #' @template verbose
-#' @param ncols Deprecated. Value is now calculated automatically.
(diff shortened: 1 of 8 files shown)


okenk:
This worked as I was hoping for both `SSplotComparisons()` and `SStableComparisons()`. Thank you!

Issue #750, opened 2022-10-05
[Feature]: modify `SSplotProfile()` to show total likelihood with and without contributions from a prior


iantaylor-NOAA:
### Describe the feature request.

Sometimes folks want to show the influence of a prior on the parameter being profiled over, and other times they want to show what the data support in the absence of a prior. The discussion at https://github.com/pfmc-assessments/nwfscDiag/issues/6 concluded that a win-win solution would be to modify the `SSplotProfile()` function to show separate lines for the total likelihood with and without the prior on the parameter being profiled over (when a prior is present). This just requires grabbing the prior contribution from that parameter and subtracting it from the total and shouldn't be too hard.

### Describe alternatives you have considered

Continue to debate whether to include the prior contribution or not without reaching any resolution.

### Describe if this is needed for a management application

It may be useful for assessment reports presented to the Pacific Fishery Management Council. We can include a switch to turn off the option in case it's distracting or confusing.

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/r4ss/r4ss#code-of-conduct)

iantaylor-NOAA:
I made some progress on this issue in [profile_prior](https://github.com/r4ss/r4ss/compare/profile_prior) branch where the key changes are
* adding `par_prior_likes` to `SSsummarize()` (along with similar existing tables `pars`, `parsSD`, and `parphases`
* adding an argument `add_no_prior_line` (default = FALSE) to `SSplotProfile()` which adds a line as shown in the figure below.

Feel free to suggest any changes in the design of these additions. 

Remaining steps are the following:
* set `add_no_prior_line <- FALSE` if there's no prior on the parameter being profiled, regardless of user input.
* copy the changes from `SSplotProfile()` to `PinerPlot()` (merging these functions to avoid redundant code is sadly not something I have time or energy for right now--see #670).
![profile_plot_likelihood](https://user-images.githubusercontent.com/4992918/230947549-fec39078-363b-4bbd-b978-057ebd7e75c6.png)

PR #1060, opened 2026-01-20
Refactor YPR_timeseries function to use SPR or YPR

e-perl-NOAA:
See SS3 PR [#750](https://github.com/nmfs-ost/ss3-source-code/pull/750) for change in spr series to use "SPR" instead of "YPR".

Add ability to find "SPR" in table while retaining backwards compatibility with "YPR"

### DIFF
--- a/R/SSplotYield.R
+++ b/R/SSplotYield.R
@@ -366,7 +366,7 @@ SSplotYield <-
         col = "blue",
         yaxs = "i"
       )
-    } # end YPR_timeseries function# end YPR_timeseries function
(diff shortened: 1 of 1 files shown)


iantaylor-NOAA:
@e-perl-NOAA, thanks for working on this. Having looked through the current output in the now-merged changes associated with  https://github.com/nmfs-ost/ss3-source-code/pull/750, I think that r4ss is doing alright. The $sprseries table based on SPR_SERIES includes both YPR and SPR but they are providing different things.

PR #1081, opened 2026-06-18
Create SS_add_parameter_line() function

okenk:
This pull request does two main things:
1. Creates an SS_add_parameter_line() function that will copy a parameter line, edit user-selected columns, and then put it in a location selected by the user
2. Adds the use of the new function to the r4ss-intro-vignette

I think I also have a different version of roxygen so a bunch of the other documentation changed very very slightly.


iantaylor-NOAA:
@okenk, thanks for adding this new function!

PR #1087, opened 2026-07-27
Make retro() automatically adjust the range of years for several quantities

### LINKED ISSUE #245: SS_doReto: Add option to adjust the late recruitment start year
It would be nice if there was an option to change the late recruitment estimation year to change according to the retrospective year.

iantaylor-NOAA:
This is a long-overdue attempt to adjust settings related to recruitment, blocks, and forecasts in retrospectives, where the range of years where things are taking place should be adjusted to reflect the peeling away of recent data. This should fix #245 (also discussed in the SS3 issue https://github.com/nmfs-ost/ss3-source-code/issues/35). Thanks to @zandjyo for the inspiration to work on this now.

The changes are implemented via a new argument in r4ss::retro(): 
`adjustments = c("recdevs", "biasadj", "blocks", "forecast", "benchmarks")`
where the inclusion of each element causes the following changes to the model:
- "recdevs": the ending year of the main recruitment period is shifted back by the number of retrospective years
- "biasadj": then the 2 values controlling the descending part of the bias adjustment ramp are similarly adjusted
- "blocks": the blocks are checked and any block which is used by at least one parameter and spans a range that only includes years with the data peeled away is removed from the block design and any time-varying parameters associated with that block are removed from the model
- "forecast": the "Fcast_years" over which averages are calculated for forecast quantities are shifted
- "benchmarks" the "Bmark_years" over which the averages are calculated for benchmarks are shifted

For the last 2 on that list (forecast and benchmarks changes), there's a check to see if the start year of the interval is closer to the start or end year of the model. If it's closer to the start year (e.g. your interval is 1916 to 2025), only the recent value is not changed. If the intervals is something like the most recent 5 years (e.g. entered as -4 to 0), then both values are changed so the new intervals remains to correspond to the most recent 5 years of data that haven't been peeled away.

By default all the changes are turned on so I changed the default for `verbose` to TRUE, so you will see lots of messages, like those pasted below. This is probably too much information, but at this stage while the feature is experimental, I think it's worth being really transparent about the changes. The old behavior of the function can be implemented via `adjustments = NULL`.

Initially I wrote this with lots of loops and if statements but then asked @copilot to help me functionalize the changes which I think results in a cleaner, easier to read code, but be aware that there are lots of subfunctions at the top of retro which do the work of changing the different types of model inputs.

If anyone, especially @zandjyo, has time to try this out, please install via a command like `remotes::install_github("r4ss/r4ss@retro-changes")` and let me know if it works for your model and you get reasonable results.

If anyone has conceptual concerns with how this is implemented, please comment here.

```
i Running retrospective in inst/extdata/models/widow2025//retrospectives/retro-5
i Adjusting MainRdevYrLast from 2020 to 2015
i Adjusting last_yr_fullbias_adj from 2016.2748 to 2011.2748
i Adjusting first_recent_yr_nobias_adj from 2022.6996 to 2017.6996
i Adjusting blocks for retrospective:
i Block designs referenced by parameter lines in control file: 4, 7, 9, 10, 11
i Block_Design 4 with 1 block: no change needed for retro end year 2019.
i Block_Design 7 with 3 blocks: no change needed for retro end year 2019.
i Block_Design 9 with 1 block: no change needed for retro end year 2019.
i Block_Design 10 with 1 block: no change needed for retro end year 2019.
i Block_Design 11 with 1 block: no change needed for retro end year 2019.
i Adjusting Fcast_years for retrospective:
* Fcast_years [1, 'st_year'] from -4 to -9
* Fcast_years [1, 'end_year'] from 0 to -5
* Fcast_years [2, 'st_year'] from -4 to -9
* Fcast_years [2, 'end_year'] from 0 to -5
* Fcast_years [3, 'st_year'] from -4 to -9
* Fcast_years [3, 'end_year'] from 0 to -5
i Adjusting Bmark_years for retrospective:
* Bmark_years beg_bio from 0 to -5
* Bmark_years end_bio from 0 to -5
* Bmark_years beg_selex from 0 to -5
* Bmark_years end_selex from 0 to -5
* Bmark_years beg_relF from 0 to -5
* Bmark_years end_relF from 0 to -5
* Bmark_years end_recr_dist from 0 to -5
* Bmark_years end_SRparm from 0 to -5
```


R/retro.R
```diff
@@ -108,8 +108,21 @@ retro <- function(
...
+  )
+  if (is.null(adjustments)) {
+    adjustments <- character(0)
+  } else {
+    adjustments <- match.arg(
+      adjustments,
+      choices = allowed_adjustments,
+      several.ok = TRUE
+    )
   }
```
okenk, line 150: This will not return an error if someone enters, say, `c('redcevs', 'biasadj')` because `biasadj` is allowed, but only the bias adjustment will get adjusted, not the recdev periods.

R/retro.R
```diff
@@ -121,46 +134,411 @@ retro <- function(
...
+      "SR_parms",
+      "Q_parms",
+      "size_selex_parms",
+      "age_selex_parms"
+    )
+    for (table_name in parm_tables) {
+      ctl[[table_name]][["Block"]][
+        ctl[[table_name]][["Block"]] == block_design
+      ] <- 0
+    }
```
okenk, line 264: Do you also need to change Block_Fxn to 0? I'm not sure.

R/retro.R
```diff
@@ -121,46 +134,411 @@ retro <- function(
...
+      "size_selex_parms_tv",
+      "age_selex_parms_tv"
+    )
+    if (length(removed_start_years) == 0) {
+      return(ctl)
+    }
+
+    year_pattern <- paste0(removed_start_years, collapse = "|")
+    row_pattern <- paste0("_BLK", block_design, ".*_((", year_pattern, "))$")
+
```
okenk, line 291: Maybe a brief comment explaining these regular expressions?

R/retro.R
```diff
@@ -121,46 +134,411 @@ retro <- function(
...
+      n_removed <- sum(!keep_rows)
+      if (n_removed > 0) {
+        ctl[[table_name]] <- ctl[[table_name]][keep_rows, , drop = FALSE]
+        if (verbose) {
+          cli::cli_alert_info(
+            "Removed {n_removed} time-varying row{?s} from {table_name} for Block_Design {block_design} start year{?s} {toString(removed_start_years)}."
+          )
+        }
+      }
+    }
```
okenk, line 306: Based on the if statment directly above, I am guessing nothing bad happens if the resulting data frame has zero rows?

R/retro.R
```diff
@@ -121,46 +134,411 @@ retro <- function(
...
+    if ("forecast" %in% adjustments) {
+      #_Fcast_years for averaging:  beg_selex, end_selex, beg_relF, end_relF, beg_mean recruits, end_recruits  (enter actual year, or values of 0 or -integer to be rel. endyr)
+      Fcast_years_names <- c(
+        "beg_selex",
+        "end_selex",
+        "beg_relF",
+        "end_relF",
+        "beg_mean_recruits",
+        "end_recruits"
+      )
```
okenk, line 487: Noting inconsistency in terminology (and in comment above). Not sure if it is intentional.

R/retro.R
iantaylor-NOAA: good catch! looks like changing to `several.ok = "all"` fixes this

R/retro.R
iantaylor-NOAA: good idea, adding now as 
```
    # Match tv row names tied to this block design and ending in one of the
    # removed block start years 
    # (e.g., the "BLK4..._1916" part of "Size_DblN_peak_BottomTrawl(1)_BLK4repl_1916").
```

R/retro.R
iantaylor-NOAA: These were just copied from the SS3 comments, but in both SS3 and this function they are just information and aren't used to match anything so I think it makes sense to leave consistent with SS3 rather than internally consistent.

R/retro.R
```diff
@@ -121,46 +134,411 @@ retro <- function(
...
+          "Adjusting {field} from {inputs[['ctl']][[field]]} to {inputs[['ctl']][[field]] + delta}"
+        )
+      }
+      inputs[["ctl"]][[field]] <- inputs[["ctl"]][[field]] + delta
+    }
+    inputs
+  }
+
+  # Apply the shared interval-shift rule used for vector year pairs:
+  # shift only end year for early intervals, otherwise shift start and end.
```
okenk, line 182: This is a somewhat arbitrary (but reasonable) rule where "early" is defined as starting no more than halfway into the model period. (I think I am interpereting this right?) It might be useful to describe this assumption in the function documention.

R/retro.R
iantaylor-NOAA: Just checked and it looks like it's fine. We could also change Block_Fxn to 0, but in general, folks probably aren't spending much time with the input files for the retro models, and perhaps there's value in being able to see which rows used to have blocks.

R/retro.R
iantaylor-NOAA: yes, if all the blocks are removed from a parameter type then the write function will ignore the associated table of short time-varying parameter lines so it won't matter if it's empty or not.

R/retro.R
iantaylor-NOAA: Yes, good point! I've added an @details section to the documentation which spells out that arbitrary rule along with other points related to what exactly gets changed by all the adjustment inputs

e-perl-NOAA:
@iantaylor-NOAA, do you think it's worth adding something in the vignettes showing these new changes in action?

iantaylor-NOAA:
@e-perl-NOAA, good idea for a vignette. I've just run an 8-year profile on the petrale sole assessment with and without the new adjustments (which took a couple hours because I forgot to run in parallel). I'll take a look at the results later today to see if they make sense and are worth including in a vignette.

iantaylor-NOAA:
@e-perl-NOAA, the results of explorations for a couple models showed that these adjustments have only a subtle impact on model results if you focus on the years that haven't been peeled away (as we typically do). It should avoid SS3 stopping with an error as discussed in https://github.com/nmfs-ost/ss3-source-code/issues/35 but I don't think it warrants a vignette. If folks express run into problems or confusion about how it's working for their models, we can revisit that idea.

PR #1088, opened 2026-07-31
change cli_inform() to cli_alert_*() functions

iantaylor-NOAA:
I had incorrectly concluding that `cli::cli_inform()` was preferred (perhaps based on issues like https://github.com/r-lib/cli/issues/715), but I like the different symbols associated with `cli::cli_alert_info()`, `cli::cli_alert_success()` and `cli::cli_alert_warning()` so have changed everything over in this PR. The choice of what is a warning vs info vs success is subjective, so please suggest changes if anything looks wrong to you.


e-perl-NOAA:
This seems logical to me!

PR #1092, opened 2026-08-03
make Bratio y-axis labels smarter in SSplotComparisons()

### LINKED ISSUE #1032: Support more Bratio options in SSplotComparison function
Did you carry through this change to the SSplotComparison function as well?

_Originally posted by @chantelwetzel-noaa in https://github.com/r4ss/r4ss/issues/1028#issuecomment-3049253044_

I didn't think to do this but it's a good idea. I think I'm now 90% of the way there but the old code is messy so and it's taking too long so I'm posting this issue as a reminder to circle back to it, hopefully later today.

iantaylor-NOAA:
Fixes #1032 
Thanks @akatan999 for reminding me that this wasn't complete.

Folks can test the changes before it gets merged by installing the branch via a command like `remotes::install_github("r4ss/r4ss@Bratio")`

### DIFF
--- a/R/SSplotComparisons.R
+++ b/R/SSplotComparisons.R
@@ -377,6 +377,7 @@ SSplotComparisons <-
     upperCI <- summaryoutput[["upperCI"]]
     SpawnOutputUnits <- summaryoutput[["SpawnOutputUnits"]]
     SpawnOutputLabels <- summaryoutput[["SpawnOutputLabels"]]
+    BratioLabels <- summaryoutput[["BratioLabels"]]
     btargs <- summaryoutput[["btargs"]]
     minbthreshs <- summaryoutput[["minbthreshs"]]
     sprtargs <- summaryoutput[["sprtargs"]]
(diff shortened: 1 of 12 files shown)


R/SSplotComparisons.R
```diff
@@ -2090,7 +2140,7 @@ SSplotComparisons <-
 
       abline(h = 1, v = 1, col = "grey", lty = 2)
 
-      if (btarg > 0) {
+      if (isTRUE(BratioLabel == "B/B_0") && btarg > 0) {
```
e-perl-NOAA, line 2141: @iantaylor-NOAA is the isTRUE() needed? I thought you could write `if (BratioLabel == "B/B_0" && btarg > 0)`

R/SSplotComparisons.R
```diff
@@ -411,32 +412,81 @@ SSplotComparisons <-
...
       all(is.na(SpawnOutputUnits)) ||
         any(SpawnOutputUnits == "numbers", na.rm = TRUE)
     ) {
+      # If units are numeric output (not biomass), update default Bratio text.
       labels[3] <- gsub("spawning biomass", "spawning output", labels[3])
     }
+    # Build the Bratio axis label:
+    # - B/B_0 keeps the standard fraction-of-unfished wording.
+    # - Alternative denominators show "Relative spawning biomass/output: <label>".
+    if (isTRUE(!is.na(BratioLabel) && BratioLabel != "B/B_0")) {
```
e-perl-NOAA: There are a lot of these isTRUE() statements that I think could/should be removed.

iantaylor-NOAA: @e-perl-NOAA, thank you for looking this over. The `isTRUE()` was added by copilot and I didn't really did into the reasons, so I appreciate your comment. I found this blog post about it https://win-vector.com/2018/06/11/r-tip-use-istrue/. Since `isTRUE()` protects against both NULL and NA, it looks like you could get the same functionality via either
`if (!is.null(BratioLabel) && !is.na(BratioLabel) && BratioLabel != "B/B_0"))`
or
`if (isTRUE(BratioLabel != "B/B_0"))` 
That second option seems shorter, so I'm going to explore what it would look like to similarly clean up other logical tests throughout r4ss.

iantaylor-NOAA: I went ahead and pushed some cleanup of 14 logical statements in c672cc6 (again with help from copilot). I'm not used to seeing `isTRUE()` in the code so it still looks a little funny, but the resulting code is both shorter and more robust, so I think this is the right choice.

e-perl-NOAA: That makes sense, I obviously didn't dive deeper into isTRUE, I'm just not used to seeing it so I was confused by copilot using it, especially since it's a toss up whether copilot is using best practices or not (e.g. copilot sometimes still using the `{magrittr}` pipe instead of the native pipe).

Issue #1097, opened 2026-09-03
[Bug]: {r4ss} in the package DESCRIPTION generates a "Lost braces.." note


cstawitz:
### Describe the bug

Hi! We are at the r-project sprint 2026 in Birmingham, UK, working on r-devel/r-dev-day#110.

This package is affected by the "Lost braces" note (see https://cran.r-project.org/web/checks/check_results_r4ss.html).

Since the package help file is generated from `DESCRIPTION`, instead of `r4ss-package.R` if you replace `{r4ss}` on line 51 of the DESCRIPTION file with single quotes instead of curly brackets e.g. `'r4ss'` it should resolve the CRAN note. However, this is a bug in `roxygen2` so if you felt strongly about including the curly braces, you could instead file a bug with `roxygen2` to support the use of `{}` instead. 

### To Reproduce

1. clone the repo
2. run `tools::checkRd("./man/r4ss-package.Rd")`

### Expected behavior

This should run cleanly without a `NOTE`

Issue #710, opened 2022-07-14
Next CRAN submission


iantaylor-NOAA:
Timing could align with next release of Stock Synthesis or a major change in r4ss. 

Checklist update from previous submission discussed in #206, where example models, tests, and vignette are now small enough to fit within the 5 MB limit so no longer need to be excluded.

- [ ] re-read http://r-pkgs.had.co.nz/release.html#release-submission
- [ ] update `NEWS.md`
- [ ] update `cran-comments.md`
- [ ] use `devtools::release()` which asks for confirmation of the following checks
  - [ ] `devtools::spell_check()` # will be faster now that `inst/WORDLIST has been created
  - [ ] `devtools::check()`
  - [ ] `devtools::check_rhub()`
  - [ ] `devtools::check_win_devel()` # identified some issues with broken URLs
  - [ ] `revdepcheck::revdep_check()` # checks ss3sim and SSMSE
- [ ] confirm the submission to CRAN via link in automated email after uploading
- [ ] expect possible issues associated with example models, tests, and vignette
- [ ] merge `CRAN` branch (if used for submission)
- [ ] create a github release which matches the CRAN submission

iantaylor-NOAA:
It could be useful to update the CRAN version of r4ss to work with the 3.30.20 release. However, I'm uncertain on timing. The recommendation is to not update CRAN too often (from https://cran.r-project.org/web/packages/policies.html#Submission):
> Submitting updates should be done responsibly and with respect for the volunteers’ time. Once a package is established (which may take several rounds), “no more than every 1–2 months” seems appropriate. 

Therefore, would it make sense to wait until more folks have had a chance to try {r4ss} with the new SS3 release to fix any of the bugs that will inevitably be discovered before submitting to CRAN, or just trust that the testing has been adequate and most models will work so it's better to get it submitted sooner rather than later?

k-doering-NOAA:
I like the idea of waiting a week or two in case bugs are found and reported, but only if that works for your schedule, @iantaylor-NOAA !

iantaylor-NOAA:
We decided to wait a week or two and now it's been 3 months.
Problems like those faced in #756 could have been avoided with a CRAN update. 

Getting this done quickly (probably next week as I don't have time this week) would avoid having to submit another version too soon after that to go with the future Stock Synthesis 3.30.21 release.

iantaylor-NOAA:
[...]

PR #1098, opened 2026-09-03
create skeleton package for CRAN

iantaylor-NOAA:
This is a package skeleton which will never be merged, but provides a way to update the long-out-of-date CRAN repo with a redirect to the version here on github. I suggested this idea 2 years ago in #710 but at the time it seemed like too much work. Now github copilot has done most of the work so it seems worthwhile.

Hat tip to @cstawitz for #1097 which reminded me of the existence of the CRAN version which this branch will replace. I'm opening the draft PR just to confirm that the skeleton package still passes checks before submitting to CRAN.


cstawitz:
Thanks @iantaylor-NOAA ! FWIW, this came up during a bug bash where we focused on the packages on CRAN that produce this NOTE in priority order of most downloads, and this was about 70th on the list of over 300. So you should be proud `r4ss` is still very popular! Most of the packages have not been updated in many years, and nearly all of the maintainers I've contacted have been surprised at the level of use they're still getting. The CRAN version of r4ss was downloaded 506 times last month. (you can check [here](https://hadley.shinyapps.io/cran-downloads/))  So I would definitely recommend archiving or updating the CRAN version if people shouldn't be using the version on there!

iantaylor-NOAA:
> Thanks @iantaylor-NOAA ! FWIW, this came up during a bug bash where we focused on the packages on CRAN that produce this NOTE in priority order of most downloads, and this was about 70th on the list of over 300. So you should be proud `r4ss` is still very popular! Most of the packages have not been updated in many years, and nearly all of the maintainers I've contacted have been surprised at the level of use they're still getting. The CRAN version of r4ss was downloaded 506 times last month. (you can check [here](https://hadley.shinyapps.io/cran-downloads/)) So I would definitely recommend archiving or updating the CRAN version if people shouldn't be using the version on there!

@cstawitz, thansk for the info. I've been trying to update the CRAN version but I'm not having any luck with `devtools::submit_cran()`. After running the function, I'm supposed to click a link in a confirmation email, but I never get the email in spam or my inbox, even after modifying DESCRIPTION to make sure it's obvious. As far as you've heard, have other maintainers had any problems submitting the CRAN? My next step was going to be trying the manual submission process via https://cran.r-project.org/submit.html.

PR #1099, opened 2026-09-08
add function clean_files() to remove unwanted ADMB output

iantaylor-NOAA:
I have moving around and using a .bat file on windows for many years to clean out unwanted files before copying directories of SS3 files to Google Drive or other shared spaces. However, it finally occurred to me that this functionality could be built into r4ss directly to save making copies of the .bat and make it available to everyone.


e-perl-NOAA:
@norellia-NOAA idk if this will be useful to you, I feel like you probably already have a workflow that works well enough but wanted to alert you to this new function

R/ss3_data_to_fims.R
e-perl-NOAA, line 75: If EWAA/wtatage aren't needed, does this function require the report.sso file? If not, then I would possibly think about adding that statement somewhere because I've run into that when I just wanted to do a simple conversion with files

